PR #39, opened 2020-11-17
waf-profiles-saving

Aviv-Galmidi:
* Added serialized input component
* Fixed bug preventing WAF Profiles from being saved correctly


Aviv-Galmidi:
This could not have been done using a computed getter/setter as we serialize the WAF Sigs IDs programmatically with changing parameters.
VueJS recommends using a component wrapping input for two-way filter and v-model
https://vuejs.org/v2/guide/migration.html#Two-Way-Filters-replaced

PR #52, opened 2020-12-08
tag-autocomplete-fixes

Aviv-Galmidi:
* Fixed bug where new tags wouldn't show up until input is refreshed + tests
* Fixed bug where new tags wouldn't be saved in multiple tags input in some occasions + tests
* Added insensitivity to tag autocomplete input type + tests
* Replaced all vue.js props and event names to use the recommended kebab-case


Aviv-Galmidi:
Merging a couple PRs into one branch

PR #53, opened 2020-12-08
Api validation

Aviv-Galmidi:
* Fixed error in api validation preventing the user from saving a profiling list without tags and entries
* Added validation that no entity is saved without a name


Aviv-Galmidi:
Merging a couple PRs into one branch

PR #56, opened 2020-12-14
Create run-curieconf-client-tests.yml

Aviv-Galmidi:
This action should run unit-tests of the curieconf client and display any failed tests


Aviv-Galmidi:
The action does not publish the result nor coverage of the client's unit test but it will fail if any of the tests fail it

PR #76, opened 2021-01-19
Pin nginx image to 1.19

flaper87:
The uiserver image depends on a script that exists in the builder image and that is not being copied into the final build. This PR fixes that

### DIFF
--- a/.github/workflows/run-curieconf-tests.yml
+++ b/.github/workflows/run-curieconf-tests.yml
@@ -1,5 +1,5 @@
 name: Run curieconf Python tests
-on: [push, pull_request]
+on: [push, pull_request_target]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


xavier-rbz:
Hello,

This file comes from the `nginx:latest` image (see its [Dockerfile](https://github.com/nginxinc/docker-nginx/blob/41156d8a36bd03b2fb36353ba31f16ada08d9e48/mainline/debian/Dockerfile#L106)).
I can see it present in an image I built from the current commit on branch tzurydev:
```
docker run --rm -ti --entrypoint ls curiefense/uiserver:v1.0.0-596-g3e0934b-76fc11f27356 -l /docker-entrypoint.sh
-rwxrwxr-x 1 root root 1202 Dec 15 20:20 /docker-entrypoint.sh
```

Please comment if I missed something.

Cheers

flaper87:
> This file comes from the `nginx:latest` image (see its [Dockerfile](https://github.com/nginxinc/docker-nginx/blob/41156d8a36bd03b2fb36353ba31f16ada08d9e48/mainline/debian/Dockerfile#L106)).

Ah, I see what happened. I had a local `nginx:latest` image that was 8 months all. The curiefense build process didn't trigger a pull. 

How would you feel about switching from using `nginx:latest` to pinning a specific version? Generally, that tends to make long term maintenance easier, it helps with making builds reproducible, and should avoid issues like this.

I'm happy to update this PR to ping the nginx version that is currently being used.

xavier-rbz:
Oh, that explains it then :) nice catch.
Agreed, not using `:latest` would avoid this kind of problems. And we'd know what is actually running.

flaper87:
I feel confident about this one, sending it through.

PR #77, opened 2021-01-20
Add KIBANA_URL required by the curielogger startup script

flaper87:


flaper87:
This was fixed in another branch

PR #78, opened 2021-01-20
Add CONTRIBUTING.md

jdorfman:


jdorfman:
@tzuryby I need help with the failed test.

CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+
+# 📑 Submitting a Pull Request (PR)
+
+* Fork the repo.
+
+* Create your PR. 
+
+* Tests will automatically run for you.
+
+* We will **NOT** merge any PR that is not passing tests.
```
flaper87, line 19: Wondering if we could skip this line. I think it's safe to assume this behavior. Mentioning it (with emphasis in NOT) may come across the wrong way.

CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
+👋 Hey there fellow open source contributor! Please read the following guidelines below carefully to maximize the chances of your PR being merged. ⤵️
+
+# 💬 Communication
+
+* ⚠️ **Before starting work on a major feature**, please reach out to us via [GitHub Discussions](https://github.com/curiefense/curiefense/discussions/categories/q-a), [Twitter DM](https://twitter.com/messages/compose?recipient_id=1321123258993938435),
```
flaper87, line 5: Wondering how well the `twitter DM` option has worked out so far. It's a bit uncommon to use twitter as a platform for contributors to reach out to the project. Perhaps focusing on GH Issues only would be better. It helps to keep the history in one place.

CONTRIBUTING.md
jdorfman: Good call, on it

CONTRIBUTING.md
jdorfman: Another good call. On it.

Aviv-Galmidi:
I've pulled the latest changes from master into this branch. The tests pass now

PR #79, opened 2021-01-22
Allow to use data streams for docker-compose based environments

flaper87:
Few changes here:

- Switch from using index mappings to using index_templates
- Use data streams as default but allow for disabling them on older ES deployments
- Add support for MaxMind's City database
- Adds new blob type to curiesync/confserver
- Store longitude / latitude data for requests locations
- gofmt for curielogger


flaper87:
FWIW, I don't think that failing job will pass since (IIRC) it's running from the forked repo, which doesn't have the right permissions. I may be wrong here. Regardless, we should replace that publish step with a step that works for everyone. Will take a look.

flaper87:
> FWIW, I don't think that failing job will pass since (IIRC) it's running from the forked repo, which doesn't have the right permissions. I may be wrong here. Regardless, we should replace that publish step with a step that works for everyone. Will take a look.

fixed this https://github.com/curiefense/curiefense/commit/4b24aba665d6ae17b42d79a0165cf90f98d4df84

tzuryby (approved):

PR #81, opened 2021-01-27
Configure logstash pipelines

flaper87:
This is in preparation for the work being done on ES-based metrics


xavier-rbz (commented):
Thanks for working on this! I'll update the configuration for helm environments later ([here](https://github.com/curiefense/curiefense/blob/master/deploy/curiefense-helm/curiefense/charts/curiefense-log/templates/logstash-configmap.yaml) and [here](https://github.com/curiefense/curiefense/blob/master/deploy/curiefense-helm/curiefense/charts/curiefense-log/templates/logstash-deployment.yaml))

flaper87:
> Thanks for working on this! I'll update the configuration for helm environments later ([here](https://github.com/curiefense/curiefense/blob/master/deploy/curiefense-helm/curiefense/charts/curiefense-log/templates/logstash-configmap.yaml) and [here](https://github.com/curiefense/curiefense/blob/master/deploy/curiefense-helm/curiefense/charts/curiefense-log/templates/logstash-deployment.yaml))

Ah, that's where the configs are. I can help you with this too. Maybe we could find a better place to put the dir I created and have the helm chart copy the files from there.

xavier-rbz:
Where you put them looks fine. The format is different, so I'd say duplication is fine, and only adds a reasonable amount of toil.

flaper87:
> Where you put them looks fine. The format is different, so I'd say duplication is fine, and only adds a reasonable amount of toil.

I've modified the helm chart templates. I thnk we can improve this a bit by sharing the logstash configs but let's not do that in this PR! :) 

I haven't properly tested the Helm templates, any help on that front would be nice.

PR #85, opened 2021-01-28
Add workflows for Rust

flaper87:


flaper87:
You can see the last run of this action here: https://github.com/flaper87/curiefense/runs/1783527488?check_suite_focus=true

PR #106, opened 2021-02-10
avivdev

Aviv-Galmidi:
* Fixed bug preventing the user from saving url maps data
* Fixed a visual bug where limit option and response action sometimes overflow their allocated space

Signed-off-by: Curiefense Bootstrap Script <curiefense-bootstrap@reblaze.com>


flaper87 (approved):
:spaghetti:

PR #108, opened 2021-02-11
Publish docker images on master push

flaper87:
This is the first step to keep our published images up to date. Note that the images use the name of the ref as the tag name. This means that pushes to master will result in a `master` tag name being used.

I'm unsure on whether `latest-dev` and `latest` are being used elsewhere but, truth is, they haven't been updated for a while so we may as well align the workflow now, once and for all. :smile:


xavier-rbz (approved):
Looks good! I don't think latest-dev is used anywhere.

PR #114, opened 2021-02-14
UI version 1.2.20

Aviv-Galmidi:
* Cleanup of avivdev history and squash changes into 1 commit
* Fixed bug preventing the ui from being initialized (exists at 1.2.19)
* Added several linters (eslint, ts config, stylelint) with a Husky pre-commit hook to be run before every commit
* Should block commits if any of these fail - hence they should never fail after this commit as they pass now.
* Currently missing type check for .vue files with "vti" (by vetur) as this plugin is in development
* Fixed linter errors and warnings
* Fixed bug where URL Maps do not show their connected WAF/ACL policies in the top-level table

Signed-off-by: Curiefense Bootstrap Script <curiefense-bootstrap@reblaze.com>


flaper87 (approved):
🍝

PR #115, opened 2021-02-14
UI version 1.2.21

Aviv-Galmidi:
Cleaned up DatasetsUtils.ts
Moved data from DatasetsUtils to more fitting locations
Moved toast functionality to Utils.ts
Added toast failure message when the user tries to select tag with less than 3 characters
Upped coverage threshold to 80 in statements and lines. Still at 75 in branches and functions

Bug fixes:
Raised minimum length of tags added to 2 across the system
Added seconds suffix to Ban Response Action
Added value selection to AutocompleteInput component when the user blurs the input selection
Fixed bug in Publish component displaying wrong number of buckets when default buckets point to non-existing buckets


Aviv-Galmidi:
@flaper87 Should I always squash commits for the same PR? or perhaps we can squash on merge using the Squash and Merge option GitHub provides through the merge button?

flaper87 (approved):
> @flaper87 Should I always squash commits for the same PR? or perhaps we can squash on merge using the Squash and Merge option GitHub provides through the merge button?

I tried my best to always keep the commit history as clean as possible. For projects like curiefense, I tend to prefer having 1 commit per PR. Multiple commits are fine but assuming they all provide a good commit message/information. 

For example, a commit message like `bug fixes` is not really useful for someone that is going through the commit history. For this PR, I would probably opt for `squash and merge`. Otherwise, I would have preferred the `rebase and merge` option. 

Hope the above makes sense

Aviv-Galmidi:
> > @flaper87 Should I always squash commits for the same PR? or perhaps we can squash on merge using the Squash and Merge option GitHub provides through the merge button?
> 
> I tried my best to always keep the commit history as clean as possible. For projects like curiefense, I tend to prefer having 1 commit per PR. Multiple commits are fine but assuming they all provide a good commit message/information.
> 
> For example, a commit message like `bug fixes` is not really useful for someone that is going through the commit history. For this PR, I would probably opt for `squash and merge`. Otherwise, I would have preferred the `rebase and merge` option.
> 
> Hope the above makes sense

That makes sense, I do try to keep the commit message as useful and informative as possible
Note: using multiple lines in the commit message only shows the first line in github preview, clicking it provides the full commit message. I know Atlassian's bitbucket showed those commits in a nicer way, I might need to go back into writing it in one line for gitHub
I'll try to keep future features with as few commits as possible

PR #117, opened 2021-02-15
[curielogger] Migrate Elasticsearch initialization to Go

### LINKED ISSUE #99: Migrate Elasticsearch initial configs into curielogger
Curielogger currently creates the initial configs (index templates, initial index, index patterns, etc)  for Elasticsearch and Kibana using a bash script:

https://github.com/curiefense/curiefense/blob/a922fc6d48a790dd0bd1e499477ecddbf1241d53/curiefense/images/curielogger/init/start_curielogger.sh#L10-L94

Moving these steps into curielogger itself will help to unify the configuration phase (currently different between helm, docker-compose, etc) and it would help to make it easier to implement and maintain. Instead of having separate scripts and json files in the image, we'd have everything built-into curielogger and versioned along with it.

flaper87:
This migration allows us to use Elasticsearch's go bindings to initialize the indexes, templates, etc instead of using the bash scripts. By doing this, we will be able to version the index templates, as well as manage other configuration options related to Elasticsearch and/or Logstash.

depends on https://github.com/curiefense/curiefense/pull/155
closes #99


xavier-rbz (commented):
I like this idea!

With this PR, outputs `elasticsearch` and `logstash` are enabled & configured in `curielogger.yml`. So, the `CURIELOGGER_USES_ELASTICSEARCH` environment variable could be removed from deploy/compose/docker-compose.yaml, and should probably be replaced with the variable that would have the same effect -- something like `CURIEFENSE_OUTPUTS_LOGSTASH_ENABLED=true` if I read the viper config correctly? not tested yet!

"Wiring" in the helm chart should be fixed so that the `curiefense_es_forwarder` variable does what it should.

Outputs `postgres` and `fluentd` are still enabled only using environment variables. Maybe we could make these two outputs configurable in `curielogger.yml` as well?

flaper87:
> I like this idea!
> 
> With this PR, outputs `elasticsearch` and `logstash` are enabled & configured in `curielogger.yml`. So, the `CURIELOGGER_USES_ELASTICSEARCH` environment variable could be removed from deploy/compose/docker-compose.yaml, and should probably be replaced with the variable that would have the same effect -- something like `CURIEFENSE_OUTPUTS_LOGSTASH_ENABLED=true` if I read the viper config correctly? not tested yet!

This is correct!

> "Wiring" in the helm chart should be fixed so that the `curiefense_es_forwarder` variable does what it should.

I need to test this further but it should already work... hoping to be able to test everything with the e2e CI work I'm doing but I'm afraid I may have to do a manual test for now.

> Outputs `postgres` and `fluentd` are still enabled only using environment variables. Maybe we could make these two outputs configurable in `curielogger.yml` as well?

Yeah, the plan is to, eventually, move all the outputs to use this config workflow.

flaper87:
@xavier-rbz I tested this using minikube and it is working as expected:

DeepinScreenshot_select-area_20210224101014

curielogger is creating the right Elasticsearch index templates, policies, etc.

xavier-rbz (approved):
Tested with both docker-compose & helm deployments with logstash: seems to run fine, no abnormal logs, requests are visible in kibana: looks good to me!

PR #119, opened 2021-02-17
Fix various bugs and issues in the client UI with new unit tests checking those issues

Aviv-Galmidi:
* Fix plural / singular option for Branches & Commits display in both Policies&Rules page and VersionControl + tests
* Fix Download Document downloads partial data due to lazy loading docs + tests
* Improve selection of tag when input blurred + tests
* Fix some buttons were using <a> tag instead of <button>
* Fix some elements were missing tab index and keyboard control
* Fix URL Map match input is missing the regex icon
* Fix loading tag rules list from link source does not import annotations
* Fix tag autocomplete input not correctly representing information in title. Changed from ‘Value’ to either ‘Tags separated by spaces’ or ‘Tag’ based on selectionType multiple or singular, respectively
* Fix buttons with is-light class do not have border when being focused on

### DIFF
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.21",
+  "version": "1.2.22",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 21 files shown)


flaper87 (approved):
:spaghetti:

Aviv-Galmidi:
Once this is merged I'll open a new PR for 1.2.23 with more changes. Waiting as to not cause a confusion and make sure we merge the code in the correct order

PR #142, opened 2021-02-19
build-docker-images.sh: improve status reporting

xavier-rbz:
`build-docker-images.sh` used to always exit with status code=0
Now, its exit code is set to 1 if one image failed to build or push. This will allow detecting build breakages in CI, such as with https://github.com/curiefense/curiefense/runs/1936100146#step:3:4065

Also build images when a pull request targeting the `main` branch is received.

### DIFF
--- a/.github/workflows/curiefense-publish-images.yml
+++ b/.github/workflows/curiefense-publish-images.yml
@@ -1,7 +1,7 @@
-name: Build docker images on pull request
+name: Publish docker images on push
 
 on:
-  pull_request:
+  push:
     branches:
       - main
     paths:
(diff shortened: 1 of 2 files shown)


flaper87 (changes requested):
I avoided having a job that just builds the images as that is something that will happen during the e2e ci job.

I'm planning to submit a smaller e2e PR that builds and runs the images (so no e2e) so that we can have a more meaningful test. (Will do this before Monday) 

What do you think? We could then just keep the changes to the script and thw rename of the xi workflow in this PR.

xavier-rbz:
Agreed, something that builds and runs the images would be more meaningful. I'll close this PR.

PR #144, opened 2021-02-20
build-docker-images.sh: improve status reporting

xavier-rbz:
This pull request is similar to #142, keeping only improvements to the build script, but not changes to the CI.

`build-docker-images.sh` used to always exit with status code=0
Now, its exit code is set to 1 if one image failed to build or push. This will allow detecting build breakages in CI, such as with https://github.com/curiefense/curiefense/runs/1936100146#step:3:4065


flaper87 (approved):
:spaghetti:

PR #152, opened 2021-02-20
logstash: increase initialDelaySeconds to 60s

xavier-rbz:
30s is sometimes not sufficient; logstash then gets killed in a loop because its livenessProbe fails.


flaper87 (approved):
:spaghetti:

Issue #129, opened 2021-02-18
System DB name is wrong at the side menu


Aviv-Galmidi:
In our side menu we have an item called "System DB" which leads to our databases editor
The name is misleading as the editor is not unique to system database but to every internal database we or the user want to add
The name should be something in the fashion of "Database"

Aviv-Galmidi:
Merged here: https://github.com/curiefense/curiefense/pull/154

Issue #127, opened 2021-02-18
tags are sometimes added to DB when already exist


Aviv-Galmidi:
When adding tags using autocomplete, we check every tag against the current known suggestions and add it to the DB if it does not exist
In some rare cases tags are being added to the DB even if they already exist

Aviv-Galmidi:
Merged here: https://github.com/curiefense/curiefense/pull/154

Issue #126, opened 2021-02-18
Reverting documents sometime freezes up the system


Aviv-Galmidi:
reverting a document with ResponseAction to a state with undefined action causes the system to go into a circular loop and freeze

Aviv-Galmidi:
Merged here: https://github.com/curiefense/curiefense/pull/154

Issue #125, opened 2021-02-18
Tags in TagRules editor are not being saved


Aviv-Galmidi:
When editing TagRules the "Tags" input changes are not being sent to the server and not being saved

Aviv-Galmidi:
Merged here: https://github.com/curiefense/curiefense/pull/154

PR #154, opened 2021-02-21
Fix issues in the UI in preparations for the release

Aviv-Galmidi:
* Fix tags in tag rules are not being saved. Closes PR: #125
* Fix reverting a document with ResponseAction to a state with undefined action causes the system to go into a circular loop and freeze. Closes PR: #126
* Fix tags are sometimes added to DB when already exist. Closes PR: #127
* Fix side menu item System DB name is misleading as the user is able to edit databases other than “system” in this page. Closes PR: #129 
* Upgrade package dependncies according to Synk suggestions
* Fix download document request cancels before finishing properly
* Fix URL Map does not focus correctly on entry name after forking map entry
* Fix Flow Control TTL being sent as string instead of number
* Fix Response Action displaying "seconds" suffix incorrectly for every input
* Fix AutocompleteInput not refocusing correctly on input after suggestion mouse click with "autoFocus" prop

### DIFF
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.23",
+  "version": "1.2.24",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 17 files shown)


flaper87 (approved):
:spaghetti:

PR #160, opened 2021-02-22
Improve the consistency of the ui in labels, titles, names, and multiple items input dividers

### LINKED ISSUE #130: Inputs with multiple items selection have no consistency
In our system we have several inputs which enable the user to choose multiple items
Some of them separate the items using a comma while others do so using space, we need to align them all to use the same divider and be consistent

### LINKED ISSUE #134: UI labels need to be updated
A few labels within the **Policies and Rules** UI need to be updated to be consistent with current naming conventions.

**WAF Policies**: 
— In the bottom part of the UI, the Headers/Cookies/Arguments tabs include a field named “Exclude Sig”. Its name should be “Exclude WAF Rule”.
— This field also contains a tooltip: “comma separated sig IDs”. This should be “comma separated rule IDs”.

**URL Maps**:
When you expand a path map, there are two pulldowns on the right.
— “WAF Profile” should be labeled “WAF Policy”.
— “ACL Profile” should be labeled “ACL Policy”.

### LINKED ISSUE #153: Policies & Rules Search sometimes displays duplicated values
**Describe the bug**
When searching in Policies & Rules Search page sometime duplicated connections IDs are displayed (e.g. __default__ ACL is displayed twice)

**To Reproduce**
1. Go to Policies & Rules
2. Click on URL Maps
3. Create a URL Map with 2 entries with the same ACL, WAF, or Rate Limit
4. Go to Policies & Rules Search
5. Search for the new URL Map
6. Under connections you will see the connected ACL / WAF / Rate Limit twice (as those are connected to two entries)

**Expected behavior**
Should remove duplicated values before displaying connectioncs

Aviv-Galmidi:
* Fix inputs with multiple items selection have no consistency, changed the multiple items input in waf rules to use space as dividers like other places - Closes #130 
* Fix some ui labels were not updated to the current name convention - Closes #134 
  - "Exclude Sig" was changed to “Exclude WAF Rule”
  - "sig IDs" was changed to "rule IDs"
  - "WAF Profile" was change to "WAF Policy"
  - "ACL Profile" was change to "ACL Policy"
* Fix Policies & Rules Search sometimes displays duplicated connection values. Will now filter out duplications and only display the same ID for a group (acl/waf/rate-limit/url-map) once per item - Closes #153

### DIFF
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.24",
+  "version": "1.2.25",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/curiefense/ui/package.json
+++ b/curiefense/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.24",
+  "version": "1.2.25",
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve",
--- a/curiefense/ui/src/components/TagAutocompleteInput.vue
+++ b/curiefense/ui/src/components/TagAutocompleteInput.vue
@@ -98,7 +98,7 @@ export default Vue.extend({
     },
 
     inputTitle(): string {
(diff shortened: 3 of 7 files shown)


flaper87 (approved):
:spaghetti:

Issue #146, opened 2021-02-20
[CII Best Practices] Warning flags


jdorfman:
This issue stems from #75 

If you know the answers, please reply like: 

`1 = Met`
`2 = ?`
`3 = Unmet`
etc.

![image](https://user-images.githubusercontent.com/398230/108603794-de61bf00-735e-11eb-8b67-8a5a11861b4a.png)

jdorfman:
1 = Met - Node.js; Rust has it; Python = Should add - @xavier-rbz 
2= Met
3 = ?

jdorfman:
@xavier-rbz are there any Warning Flags present or planned for the Python scripts?

i.e.

> The project MUST enable one or more compiler warning flags, a "safe" language mode, or use a separate "linter" tool to look for code quality errors or common simple mistakes, if there is at least one FLOSS tool that can implement this criterion in the selected language.

PR #163, opened 2021-02-23
Python linter

xavier-rbz:
This pull request adds automated python linting tests, showing warnings and errors. It will help solving the Python part of #146


.github/workflows/run-pylint.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - name: install pylint
+        run: pip install pylint
```
flaper87, line 14: Not to start a long discussion on linters, tools, and whatnot. Do you think we could use `black` here instead of `pylint`?

https://black.readthedocs.io/en/stable/

flaper87 (changes requested):

.github/workflows/run-pylint.yml
xavier-rbz: If I understand black correctly, it's a code formatter only, it doesn't issue warnings or errors like pylint does. The "code analysis" part of pylint is what I wanted to add with this PR.
Yes, we could decide use black to format code (my personal answer would be "why not" rather than an enthusiastic "yes").

jdorfman: @flaper87 thoughts? ^

flaper87: @xavier-rbz Correct! I guess where I'm coming from is that adopting `black` to format the code should remove the need of doing a lint check. We could just check if the code is "black'ed out".

This removes the need of installing the client, server, e2e, etc. We would have to have black (there's a container somewhere for it already) and pass it the paths we want to check.

makes sense?

flaper87: I don't think we have to block this PR on this, tho. Let's move it forward and switch to black later on

flaper87 (approved):
:spaghetti:

flaper87:
sending through, I will work on moving things to use `black` instead

PR #166, opened 2021-02-24
Disallow match conflicts for URL Maps

### LINKED ISSUE #128: Default URLMaps should be read-only
In Policies&Rules page we have a document type called URL Maps. The default URL Map should be read only and only copyable instead of editable as it is the default fallback map to use when no other match is found

### LINKED ISSUE #135: Conflicting URL Maps should be prevented
A user can create multiple URL Maps for the same path, but this should not be allowed.

Currently, a user can duplicate an existing URL Map, and edit the new Map to have a different WAF Policy or ACL Policy than the original. This creates two conflicting URL Maps for the same path.

This problem is probably the most severe when it occurs for the default path, but this situation should be prevented in general. When a user attempts to save a newly edited/created URL Map, the system should verify that it does not conflict with an existing one.

Aviv-Galmidi:
* Add check for conflicting url maps and disable save on conflicts. Conflicts are check per domain match (no duplicate domain match between any two url maps) as well as map entries name matches (no duplicate map entry match between two entries of the same url map)
* Disable editing the default URL Maps default entry `/` and default domain match `__default__`
* Add Snyk package.json update suggestions
* Add Utils function which removes the classes we add to mark an input element as invalid if present

Closes #128 
Closes #135

### DIFF
--- a/curiefense/curieconf/server/curieconf/confserver/v1/api.py
+++ b/curiefense/curieconf/server/curieconf/confserver/v1/api.py
@@ -181,6 +181,7 @@ m_document_mask = api.model('Mask for document', {
     "bypass": fields.List(fields.String()),
     "deny": fields.List(fields.String()),
     "force_deny": fields.List(fields.String()),
+    "match": fields.String(),
     '*': fields.Wildcard(fields.Raw()),
 })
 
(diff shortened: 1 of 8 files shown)


flaper87 (approved):
:spaghetti:

PR #169, opened 2021-02-25
Policies & Rules url path reset fix

### LINKED ISSUE #161: Switching branches or Reverting document in Policies & Rules causes the selected doc type and doc to reset
**Describe the bug**
Switching branches in Policies & Rules causes the selected doc type and doc to reset
Reverting a document causes the selected doc id to reset

**To Reproduce**
Steps to reproduce the behavior: (Switching branches)
1. Go to 'Policies & Rules'
2. Click on 'Document type selection dropdown' in the top left
3. Select Rate Limit
4. Click on 'Branch selection dropdown' in the top left
5. Select a different branch
6. The page will reload ACL Policies instead of the current Rate Limits
Steps to reproduce the behavior: (Revert)
1. Go to 'Policies & Rules'
2. Click on "Add new document" in the top right
3. Select the second document in the documents selection dropdown in the top right
4. Change the document a bit
5. Click on "Save" in the top right
6. Scroll down to "Version History"
7. Revert to the original state
8. The page will reload the first document instead of the current one

**Expected behavior**
The page should attempt to load the same document type with the same document id, if it fails it should attempt to load the same document type with the first document id of the list, if that also fails only then it should load the first document type (acl policies)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Chrome
 - Version 1.2.25

Aviv-Galmidi:
Change Policies & Rules to not reset the url path when switching branhes or reverting a document to a previous state

* Fix Policies & Rules reset path when switching branches
* Fix Policies & Rules reset path when reverting a document

Closes #161

### DIFF
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.26",
+  "version": "1.2.27",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/curiefense/ui/package.json
+++ b/curiefense/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.2.26",
+  "version": "1.2.27",
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve",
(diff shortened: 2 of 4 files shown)


Aviv-Galmidi:
Not sure what these failing lua/shell checks are but they don't seem related to my PR

flaper87 (approved):
:spaghetti:

flaper87:
merging as failures are unrelated

PR #172, opened 2021-02-25
helm: fix elasticsearch permissions (does not start on GKE)

xavier-rbz:


flaper87 (approved):
:spaghetti:

PR #176, opened 2021-02-26
e2e: use the 'tool publish' command & improve its documentation

### LINKED ISSUE #48: "curieconfctl sync export" should request the server to export to the bucket
`curieconfctl sync export` tries to export to the bucket locally, rather than asking the server to do so. This breaks when using docker-compose, where by default the bucket is mounted to /bucket, which does not exist on the host.

See https://github.com/curiefense/curiefense/blob/master/curiefense/curieconf/client/curieconf/cli/__init__.py#L346

-> call the export API instead of calling the cloudstorage API

xavier-rbz:
closes #48

### DIFF
--- a/curiefense/curieconf/client/curieconf/cli/__init__.py
+++ b/curiefense/curieconf/client/curieconf/cli/__init__.py
@@ -321,6 +321,10 @@ tool = typer.Typer()
 
 @tool.command("publish")
 def publish(config:str, url:str, version:str = None):
+    """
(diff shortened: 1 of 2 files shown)


flaper87 (approved):
Tests failed for me when I tested this before writing https://github.com/curiefense/curiefense/issues/48#issuecomment-786876241

I think the failures were mostly related to my environment and not the change itself. Thanks for the PR!

PR #177, opened 2021-02-26
shell scripts: fix them by following shellcheck recommendations

xavier-rbz:


deploy/curiefense-helm/deploy.sh
```diff
@@ -29,11 +29,10 @@ if ! kubectl get namespaces|grep -q curiefense; then
     echo "curiefense namespace created"
 fi
 
-helm upgrade --install --namespace curiefense --reuse-values $HELM_ARGS \
+# shellcheck disable=SC2086
+if helm upgrade --install --namespace curiefense --reuse-values ${HELM_ARGS} \
```
flaper87, line 33:
```suggestion
if ! helm upgrade --install --namespace curiefense --reuse-values ${HELM_ARGS} \
```

We only want to describe pods/fetch logs if the deployment fails. Unless I'm missing something, of course.

deploy/istio-helm/deploy.sh
```diff
@@ -29,14 +29,13 @@ if [ -n "$NOPULL" ]; then
     PARAMS+=("--set" "global.imagePullPolicy=Never")
 fi
 
-helm upgrade --install --namespace istio-system --reuse-values \
+# shellcheck disable=SC2086
+if helm upgrade --install --namespace istio-system --reuse-values \
```
flaper87, line 33: ditto

flaper87 (changes requested):

deploy/curiefense-helm/deploy.sh
xavier-rbz: nice catch. Applied in the next commit

deploy/istio-helm/deploy.sh
xavier-rbz: ditto

PR #180, opened 2021-02-27
helm: explicitely disable logstash when not needed (it is enabled by …

xavier-rbz:
…default)


flaper87:
I've created this one that will disable all loggers by default: https://github.com/curiefense/curiefense/pull/185

What do you say we go with #185 instead of #180 ?

xavier-rbz:
let's do that.

PR #182, opened 2021-02-27
helm: add variable to set XFF_TRUSTED_HOPS

xavier-rbz:


xavier-rbz:
XFF_TRUSTED_HOPS is configured in a VirtualService and not as a global, per-envoy-instance setting.

PR #194, opened 2021-03-01
control flow with peripherial matching and custom actions

tzuryby:
custom ersponse tested 
ttl expiration
HCA matching


curiefense/curielogger/convert-bq/convert.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if len(ngxjson["upstream_addr"]) > 1:
+        upstream = ngxjson["upstream_addr"].split(":")
+        addr = upstream[0]
+        # port = int(upstream[1].split(",")[0])
+        eljson["upstream"] = {
+            "remoteaddress": addr,
+            "remoteaddressport": 0
+        }
+
```
flaper87, line 73: This is missing the `geo` data that I added in the server. I can work on a PR myself to add it to the repo.

curiefense/curielogger/convert-bq/steps.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+#### step 04 - insert into ES
+
+```bash
+for FI in bulk-*.json;
+do
+    echo $FI;
+    curl -s -H "Content-Type: application/x-ndjson" -XPOST localhost:9200/_bulk --data-binary "@$FI";
+done
```
flaper87, line 38: Is there a chance for us to use curielogger directly here instead of doing the convert and then bulk insert? It looks like this is a great opportunity for dogfooding but I may be missing something.

flaper87 (commented):
none of my comments are blockers, although there are some CI failures that would be nice to fix before getting this in.

xavier-rbz (changes requested):
I'd like to postpone merging until we have passing e2e tests for this feature

flaper87:
@tzuryby is this PR still relevant? Or will it be superseded by the Rust rewrite?

Issue #190, opened 2021-02-28
Logs not sent to logstash


xavier-rbz:
Deployed from `main` (1b962bfbbb0ffe0a45a601a31bba2fd61922d50a) using helm.

Configuration looks OK, I have these variables in the curielogger container:
```
root@curielogger-6665fcd7dc-l6sfm:/# export|grep CUR
declare -x CURIELOGGER_INITIALIZE_KIBANA_INDEX="true"
declare -x CURIELOGGER_KIBANA_URL="http://kibana:5601/"
declare -x CURIELOGGER_METRICS_PROMETHEUS_ENABLED="true"
declare -x CURIELOGGER_OUTPUTS_LOGSTASH_ENABLED="true"
declare -x CURIELOGGER_OUTPUTS_LOGSTASH_URL="http://logstash:5001/"
(...)
```

Yet in the output, I cannot see debug messages related to logstash:
```
kubectl logs -n curiefense curielogger-6665fcd7dc-67l88
2021/02/28 11:48:10 Starting curielogger v0.3-dev1
2021/02/28 11:48:10 [INFO] Log level set at debug
2021/02/28 11:48:10 [INFO] Channel capacity set at 65536
2021/02/28 11:48:10 Prometheus exporter listening on :2112
2021/02/28 11:48:10 GRPC server listening on :9001
2021/02/28 11:48:10 [INFO] Prometheus (prometheus) metrics updating routine started
```

No kibana or elasticsearch initialization takes place. Since we're in debug mode, I expected to see logs from https://github.com/curiefense/curiefense/blob/main/curiefense/curielogger/curielogger/main.go#L487 https://github.com/curiefense/curiefense/blob/main/curiefense/curielogger/curielogger/main.go#L826 https://github.com/curiefense/curiefense/blob/main/curiefense/curielogger/curielogger/main.go#L834

xavier-rbz:
After adding debug code (https://github.com/xavier-rbz/curiefense/commit/c56ed5d8418e20cec08dd766b42b9ac2656c6e2d), I can see that changes to `CURIELOGGER_LOG_LEVEL` are reflected in the output, but not changes to `CURIELOGGER_OUTPUTS_LOGSTASH_ENABLED` or `CURIELOGGER_OUTPUTS_LOGSTASH_URL`.

Sample output after setting `CURIELOGGER_OUTPUTS_LOGSTASH_ENABLED=true` in `docker-compose.yml`:
```
[...]

PR #197, opened 2021-03-01
enable logstash by default until #190 is fixed

xavier-rbz:


xavier-rbz:
self-approving this quickfix...

PR #199, opened 2021-03-01
curielogger: add missing call to enable loggers

xavier-rbz:
This is still not 100% reliable: I had to restart (kill) curielogger to make it initialize kibana properly from a fresh docker-compose environment, but at least it sends logs.


xavier-rbz:
Actually the call IS present if initialization was successful, so the problem we're seeing is somewhere else.

PR #202, opened 2021-03-02
curielogger: disable retry logic

flaper87:


flaper87:
Sending through as logging to logstash is currently broken

PR #209, opened 2021-03-03
[curielogger] Add docker-compose e2e tests

flaper87:


deploy/deploy-ci-compose.sh
```diff
@@ -0,0 +1,45 @@
...
+ENVOY_UID=0
+DOCKER_TAG=$DOCKER_TAG
+
+CURIE_BUCKET_LINK=file://$WORKDIR/bucket/prod/manifest.json
+EOF
+
+cat $WORKDIR/ci-env
+
+pushd deploy/compose || exit
+dcoker-compose up -d --env-file "$WORKDIR/ci-env"
```
xavier-rbz, line 41:
```suggestion
docker-compose up -d --env-file "$WORKDIR/ci-env"
```

deploy/deploy-ci-compose.sh
```diff
@@ -0,0 +1,45 @@
...
+pushd "$ROOT_DIR" || exit
+
+pushd curiefense/images || exit
+./build-docker-images.sh
+popd || exit
+
+# curieconfctl will try to write to this
+# path during the tests. This is currently
+# not configurable.
+mkdir -p "$WORKDIR/bucket"
```
xavier-rbz, line 28: with docker-compose, `/bucket` doesn't need to be mounted on the host, the default should be fine. Unless you plan to do something with these files, this can be removed.

deploy/deploy-ci-compose.sh
```diff
@@ -0,0 +1,45 @@
...
+chmod 777 "$WORKDIR/bucket"
+
+cat <<EOF > $WORKDIR/ci-env
+ENVOY_UID=0
+DOCKER_TAG=$DOCKER_TAG
+
+CURIE_BUCKET_LINK=file://$WORKDIR/bucket/prod/manifest.json
```
xavier-rbz, line 35: Does this work? It means that containers will try to read to that location, which is not mounted; by default, the bucket is located at `/bucket` inside the containers

deploy/deploy-ci-compose.sh
```diff
@@ -0,0 +1,45 @@
...
+EOF
+
+cat $WORKDIR/ci-env
+
+pushd deploy/compose || exit
+dcoker-compose up -d --env-file "$WORKDIR/ci-env"
+
+# Some debug information
+docker-compose top
+docker-compose logs
```
xavier-rbz, line 45: maybe filter logs using `e2e/logs-smoke-test/checklogs-compose.sh`? That script removes logs we know about, that are expected. I typically use it this way:
```
cd e2e/logs-smoke-test/
./checklogs-compose.sh
grep . *log
```

xavier-rbz (changes requested):

flaper87:
@xavier-rbz and I worked on this one so, sending it through now.

PR #211, opened 2021-03-03
[curielogger] Reimplement the configure retry logic

flaper87:
With this commit, the configuration of the output will be attempted before registering the output, which helps curielogger guarantee that the output is up, and ready to receive messages

The retry logic in action on a single screenshot:

DeepinScreenshot_select-area_20210303200156


xavier-rbz (approved):
Tested with docker-compose & helm (logstash) -- looks good.

PR #212, opened 2021-03-04
[curielogger] Use `implant` to embed files

flaper87:
Move index settings, templates, patterns, etc outside the elasticsearch module and use implant instead to create a static files module to access such data.

This commit also adds Kibana's dashboard so that it's created during the curielogger configuration step


curiefense/images/curielogger/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM golang as builder
 
 COPY curielogger /curielogger
 WORKDIR /curielogger
+RUN go get -u github.com/skx/implant
+RUN implant -input files/ -output curielogger/static.go
```
xavier-rbz, line 6: `static.go` seems to be generated; if that's the case, can remove `static.go` from the repo?

xavier-rbz (approved):

curiefense/images/curielogger/Dockerfile
flaper87: I agree! I had added it before I put the generation in the image. I will remove it in a follow-up patch after documenting the process.

Another reason why I pushed it is to not force people to install `implant` on their local env (in case they want to build the binary without docker).

PR #241, opened 2021-03-14
Add autocomplete support to WAF Policies editor and resolve a bug in URL Maps editor

### LINKED ISSUE #170: UI - WAF Policy should autocomplete and suggest possible matches for WAF Rules input
Policies & Rules -> WAF Policies

The user is able to add rule entries at the bottom of the page to either Headers, Cookies, or Args
The user is able to add WAF rules IDs to be excluded from the new rule entry

Currently the user needs to type the IDs themselves, we can use our autocomplete component and add possible suggestions and autocomplete to the process from our list of WAF Rules

### LINKED ISSUE #227: UI - Reverting documents version may cause an error in linked documents
**Describe the bug**
When reverting a document of type ACL Policy / WAF Policy / Rate Limit / URL Map there is a chance the old version is not aligned with other connected documents, for example reverting the Rate Limit document may cause an existing Rate Limit to not exist anymore, but it is still connected to a URL Map

**To Reproduce**
1. Go to 'Rate Limits' editor in 'Policies & Rules'
2. Create a new 'Rate Limit'
3. Go to 'URL Maps' editor in 'Policies & Rules'
4. Open an entry map of the current 'URL Map'
5. Add the newly created 'Rate Limit' to 'Rate Limit Rules' in the bottom left corner
6. Save
7. Go back to 'Rate Limits' editor in 'Policies & Rules'
8. Using the 'Version History' at the bottom of the page, revert the document to its initial state
9. Go back to 'URL Maps' editor in 'Policies & Rules'
10. Open the entry map we've edited on steps 4-5
11. You will see the corrupted data breaks the flow of the UI

**Expected behavior**
The UI should at least not break when there is an undentified linked document in URL Maps
I am not sure what the implications are for something like that, but as there is no check in the API whether the given data is valid then I assume it will not break the system

Aviv-Galmidi:
Add autocomplete support to WAF Policies editor and resolve a bug in URL Maps editor

* Add autocomplete input in Policies & Rules -> WAF Policies -> Headers/Cookies/Arguments -> Exclude WAF Rule. Will now autocomplete and suggest possible rule IDs based on available WAF rule IDs. Closes #170
* Remove SerializedInput.vue and SerializedInput.spec.ts as they are no longer being used anywhere
* Fix bug in URL Maps where a map entry with invalid rate limit IDs would display an empty line with only "remove" button and no data. Closes #227


tzuryby (approved):
LGTM

PR #242, opened 2021-03-15
Run CI jobs only on PRs to target

flaper87:

### DIFF
--- a/.github/workflows/run-curieconf-tests.yml
+++ b/.github/workflows/run-curieconf-tests.yml
@@ -1,5 +1,10 @@
 name: Run curieconf Python tests
-on: [push, pull_request_target]
+
+on:
+  pull_request_target:
+    paths:
+      - "curiefense/curieconf/**"
+      - ".github/workflows/run-curieconf-tests.yml"
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


.github/workflows/run-curieconf-tests.yml
```diff
@@ -1,5 +1,10 @@
 name: Run curieconf Python tests
-on: [push, pull_request_target]
+
+on:
+  pull_request_target:
+    paths:
```
Aviv-Galmidi, line 5: These are the paths for the UI tests and not the Python tests. The correct paths are:
- "curiefense/curieconf/**" instead of "curiefense/ui/**"
- ".github/workflows/run-curieconf-tests.yml" instead of ".github/workflows/run-curieconf-client-tests.yml"

Aviv-Galmidi (changes requested):

.github/workflows/run-curieconf-tests.yml
flaper87: jeez, too much copy pasta :D

Aviv-Galmidi (approved):

PR #243, opened 2021-03-15
Fix e2e tests related to ratelimit events

xavier-rbz:
Related doc: https://docs.curiefense.io/settings/policies-rules/rate-limits#the-event-option-changing-the-meaning-of-the-rate-limit


flaper87 (approved):
:spaghetti:

PR #254, opened 2021-03-21
Make all e2e tests pass

xavier-rbz:
Code fixes in both tests & curieproxy code that get us to 0 failing e2e tests.

Checks added to the LUA code adds some duplication, but that could be addressed later.

Closes #254

### DIFF
--- a/.github/workflows/e2e-compose-tests.yml
+++ b/.github/workflows/e2e-compose-tests.yml
@@ -45,7 +45,7 @@ jobs:
             echo "-- Run e2e tests --"
             pushd e2e
             IP=127.0.0.1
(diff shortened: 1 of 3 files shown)


xavier-rbz:
So... e2e tests don't pass, I can see this error:
```
Error: "helm upgrade" requires 2 arguments
```

But e2e tests on docker-compose **do** pass! All tests are enabled.

Aviv-Galmidi:
> So... e2e tests don't pass on minikube, I can see this error:
> 
> ```
> Error: "helm upgrade" requires 2 arguments
> ```
> 
> But e2e tests on docker-compose **do** pass! All tests are enabled.

From the docs: `helm upgrade [RELEASE] [CHART] [flags]` source [here](https://helm.sh/docs/helm/helm_upgrade/). so helm upgrade requires release and chart arguments

tzuryby (approved):

flaper87:
This breaks CI for the minikube scenarios. I don't think we should've merged it this way.

PR #259, opened 2021-03-29
[e2e] fix arguments passed to deploy.sh. Fixes e2e tests.

xavier-rbz:
Use pull_request instead of pull_request_target:

* allows actually testing changes to the testing infra
* avoid exposing build secrets or providing code from unmerged PR with write-access to the repository

Test results from `pull_request_target` should be disregarded, as they do not reflect the state of the repository after merging this.


flaper87 (approved):
thanks, let's see if the switch to `pull_request` fixes our current CI failures :)

PR #301, opened 2021-04-03
Refactor logger

maxim-kuderko:
Highlights:
+ Separated grpc interface of the logger for the logger logic
+ Rewrote the loggers to implement an io.WriteCloser interface to offer better compatibility
+ Made Viper be the only way to get configs inside the app to avoid any confusion, moving forward we should only use ENV vars to change settings of the app
+ Added an Stdout driver
+ Added a "bucket" driver to support flushing logs directly into cloud storage/files + compressing on the fly
+ deprecated the logstash and fluentd drivers, moving forward the stdout should be used with any existing log collection util in the environment
+ changed internal logs to be flushed to stderr and now it's using logrus instead of the default log pkg

Todo before review:
+ add a few tests (**ideas are appreciated as there is very little logic in the service**)
+ move the log schema in ES/logstash/fluent to a separate folder with maybe a single line install bash script, the logger shouldn't re configure ES on every boot & like stated before is should not be aware of any logging collectors in the future - **Maybe even a plain "telementry" folder to hold all configs of external services like grafana, es, logstash fluent etc... WDYT?**
+ some more testing required


flaper87:
Hey @maxim-kuderko 

thanks for the PR and for cleaning up the curielogger implementation. It was due for a good refactor. 

A few things came to mind from the PR description that I think would be worth discussing before diving into the reviews:

> Made Viper be the only way to get configs inside the app to avoid any confusion, moving forward we should only use ENV vars to change settings of the app

I don't mind allowing the use of environment variables to set some settings. However, I do think we should keep the ability to use a configuration file too. Some loggers may require more complex settings and, for some of these cases, using a configuration file is ideal. Environment variables are common in a cloud-native environment but I think it would be a mistake for us to assume that it will always be run in such an environment

I'd prefer to create an issue for this refactor so we can have a longer discussion about the changes here. Touching this part will have a great impact in the user experience.

> Added an Stdout driver
> Added a "bucket" driver to support flushing logs directly into cloud storage/files + compressing on the fly

What do you say we create a separate issue for each of these and pull them into their own PRs?

> deprecated the logstash and fluentd drivers, moving forward the stdout should be used with any existing log collection util in the environment

Let's hold-off on removing any output logger for now. I would like for us to properly discuss this, evaluate pros/cons and eventually make a decision. These loggers were released in the `1.3.0` and, for better or for worse, we have to provide support for them and, at the very least, a deprecation path. 

Please, create an issue for this too so we can discuss this further :hugs: 

> Rewrote the loggers to implement an io.WriteCloser interface to offer better compatibility

Nice! :+1: 

> Separated grpc interface of the logger for the logger logic

I'm quite excited about this one. Few comments here:

1. It would be great to have a separate issue for it, describing what the goal is, and how we intend to structure the code to accommodate for new HTTP proxies in the future.
2. I know you've said that this is most of the work but I would really really really prefer this to be in a separate PR! Pretty please :pray: 
3. I'm super duper excited about pulling this code into its own module

> add a few tests (ideas are appreciated as there is very little logic in the service)

We mentioned in slack that we could use [ginkgo](https://onsi.github.io/ginkgo/). As for what to test, here are a couple of ideas:

- input (just grpc for now) -> log obect -> logger output
- Configuration loading (from file and environment variables)
- Ability to have more than one logger enabled
- Elasticsearch logger could use some tests for the ILM, indexes, policies, and other settings

the above are some ideas, I'm not suggesting they should all go into this (or into a single) PR :100: 

Thank you for looking into curielogger and questioning all the decisions made so far. I've been dreaming on giving it a more go-like look to it.

flaper87 (changes requested):
forgot to set the status when I submitted my previous comment :)

maxim-kuderko:
@flaper87 Thank you for your review :)
I've opened the following issues to continue the discussion we started at the office
https://github.com/curiefense/curiefense/issues/303
https://github.com/curiefense/curiefense/issues/304

As for the GRPC separation, implementing it in a separate PR will more or less result in the same code, I needed to encapsulate the logger in it's own pkg, create a DI framework, separate all the entities.
The only thing I might not need to do is refactor the loggers to be interfaces but it's really not that big of a change considering the entire refactor.
I've added to this PR the previous loggers, still fully functioning we can decide if to drop them in [this issue](https://github.com/curiefense/curiefense/issues/303) but for this PR they will stay regardless.

No loggers were harmed in this PR :)


Thanks for the tests suggestion, Working on them now.

flaper87:
@maxim-kuderko thanks for creating those issues!

I have a few more high-level comments before diving into the code itself:

1. Could you please use `git mv` as opposed to deleting the files and then adding new ones? The PR, as-is, will break the git-history, which is something I would like to preserve. I noticed that you did use `git mv` for some of them but the drivers' and the config are missing.
2. Please restore the configuration layer as it was. 
3. Please, restore the initialization steps for the various drivers. These are important to guarantee an optimal setup for Curiefense to write to the drivers. In the Elasticsearch case, for example, Curielogger initializes the indexes, templates, and it sets up the dashboards if Kibana is enabled.
4. What do you say we add the `tee` and `stdout` drivers in a separate PR? I'd prefer to keep this PR focused on the refactor and not adding new functionality.
5. What do you say we rename `drivers` to `outputs`. This will make things consistent with the way they are and I believe it represents better what this layer does. It outputs data to various storage technologies as opposed to using them for the lifecycle of the application.

> As for the GRPC separation, implementing it in a separate PR will more or less result in the same code, I needed to encapsulate the logger in its own pkg, create a DI framework, separate all the entities.

- Can this be done without touching the config module and the existing outputs? 
- Another option would be to modify the drivers' APIs before changing the gRPC layer. This would help us make progress in a more controllable and "testable" way.

Thanks a bunch for working on this and for your patience. Hope you understand where I'm coming from with all these comments and my obsession with making the PR smaller. It's not just about the size of the PR but also about the changes it introduces. Until very recently, there were 0 integration tests in this repo and every change introduced quite a few breakages that @xavier-rbz and I had to manually test, hence the slow changes and refactors and why I'm being as cautious as I am.

maxim-kuderko:
Done :)

Re-added all the old config files but using go:embed now :package: 
kibana ES are configured as expected

flaper87:
The fialure looks unrelated (luarocks strikes back). I pushed a new commit to fix the index mapping, which I think I broke in a previous PR. Everything seems to work as expected:

DeepinScreenshot_select-area_20210413163919
DeepinScreenshot_select-area_20210413163907
DeepinScreenshot_select-area_20210413163842


Here's the datastream after logs were sent to Elasticsearch using Logstash:

DeepinScreenshot_select-area_20210413163851

Tested the above using the `docker-compose` deployment

flaper87:
Thanks for the patience and the work here, @maxim-kuderko :muscle:

PR #307, opened 2021-04-05
curielogger: cleanup duplicated fields

flaper87:

### DIFF
--- a/curiefense/curielogger/curielogger/common.go
+++ b/curiefense/curielogger/curielogger/common.go
@@ -87,6 +87,8 @@ type NameValue struct {
 }
 
 type RequestData struct {
+	RequestId    string                 `json:"requestid"`
+	Scheme       string                 `json:"scheme"`
 	BodyBytes    uint64                 `json:"bodybytes"`
 	HeadersBytes uint64                 `json:"headersbytes"`
 	OriginalPath string                 `json:"originalpath"`
@@ -123,13 +125,8 @@ type MetadataData struct {
 }
 
 type CuriefenseLog struct {
-	RequestId string `json:"requestid"`
 	Timestamp string `json:"timestamp"`
-	Scheme    string `json:"scheme"`
 	Authority string `json:"authority"`
-	Port      uint32 `json:"port"`
-	Method    string `json:"method"`
-	Path      string `json:"path"`
 
 	Blocked     bool                   `json:"blocked"`
 	BlockReason map[string]interface{} `json:"block_reason"`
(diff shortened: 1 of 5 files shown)


e2e/test_e2e.py
```diff
@@ -162,26 +162,13 @@ def target(request):
...
-                "statement": (
-                    "SELECT Path FROM logs " "ORDER BY StartTime DESC LIMIT 1024"
-                ),
-                "parameters": [],
-            }
-            res = requests.post(self._base_url + "/logs/api/v1/exec/", json=data)
-            for log in res.json():
-                if pattern in log[0]:
-                    return True
-        else:
```
flaper87, line 179: @xavier-rbz Do you know/remember why this was needed? I would rather just use ES' DSL if this SQL query is not required :)

Aviv-Galmidi: The path "/logs/api/v1/exec/" was the old API path we used for requesting logs before we switched over to ELK and it is no longer in use, if that helps you in any way :)

curiefense/curielogger/curielogger/main.go
```diff
@@ -171,9 +165,10 @@ func (s grpcServerParams) StreamAccessLogs(x als.AccessLogService_StreamAccessLo
 					Version:     tls.GetTlsVersion().String(),
 				},
 				Request: RequestData{
+					RequestId:    req.GetRequestId(),
+					Scheme:       req.GetScheme(),
 					BodyBytes:    req.GetRequestBodyBytes(),
 					HeadersBytes: req.GetRequestHeadersBytes(),
-					OriginalPath: req.GetOriginalPath(),
```
xavier-rbz, line 176: shouldn't `OriginalPath` be removed from common.go?

curiefense/curielogger/curielogger/common.go
```diff
@@ -123,13 +125,8 @@ type MetadataData struct {
 }
 
 type CuriefenseLog struct {
-	RequestId string `json:"requestid"`
 	Timestamp string `json:"timestamp"`
-	Scheme    string `json:"scheme"`
 	Authority string `json:"authority"`
-	Port      uint32 `json:"port"`
-	Method    string `json:"method"`
-	Path      string `json:"path"`
```
xavier-rbz, line 130: May I suggest updating `es_index_template.json` to reflect these changes?

xavier-rbz (changes requested):

curiefense/curielogger/curielogger/common.go
flaper87: > May I suggest updating es_index_template.json to reflect these changes?

Oh, man! I can't believe I forgot to do this :facepalm:

xavier-rbz (approved):

Issue #317, opened 2021-04-13
Deprecate existing outputs in favor of a single one


flaper87:
Curielogger has evolved from being a logging tool for the Curiefense task into managing logs, metrics, and other events for multiple technologies. This has expanded the list of supported outputs to Logstash, FluentD, and Elasticsearch. The trend here will take Curielogger from its original purpose into a general-purpose log collector.

To avoid getting into a realm of problems that have been solved elsewhere, it was proposed and discussed to remove the support for the existing loggers in favor of a single `Stdout` one. This will leave the decision of the storage technology to the operator, rather than having Curielogger dictate what such technology should be.

Work for `1.4.0`
==

- [ ] Document the current Elasticsearch storage and configurations (ILM Policy, Index Template, Dashboard, etc)
- [ ] Document a deployment using FluentD to collect logs from stdout to send them to Elasticsearch
- [ ] Document a deployment based on filebeat to collect logs and send them to Elasticsearch
- [ ] Create a contrib dir and move the existing Elasticsearch configs there.
- [ ] Mark the existing outputs as deprecated and log a warning during the initialization steps

Deprecation Strategy
==

The current loggers will be marked as deprecated and then removed in `1.5.0`. This will give enough time for existing deployments to migrate to a configuration that doesn't depend on Curielogger talking to the storage directly.

Related discussion here: https://github.com/curiefense/curiefense/discussions/306#discussioncomment-604306

flaper87:
@maxim-kuderko let me know if I forgot something, feel free to add/modify as needed

PR #318, opened 2021-04-16
Mark all outputs as deprecated

flaper87:
As part of #317 all outputs will be deprecated and the default one will be the stdout output.

related to #317


flaper87:
Failure is unrelated and this is just adding some Warning messages referring to the deprecation of the outputs. I will get this in now.

PR #311, opened 2021-04-05
curieproxy-image: use binaries built from curiefense/envoy

xavier-rbz:
This PR:

* Removes `envoy` binaries that used to be stored in this repo.
* Updates istio to 1.9.2, and uses the right corresponding `envoy` binary.

Binaries for the `curieproxy-envoy` image are now built from [the curiefense/envoy repository](https://github.com/curiefense/envoy/blob/0250ad715e8c0463d0d4389cbfab340bc7dd7587/.github/workflows/build-envoy-for-curiefense.yml)

Binaries for the `curieproxy-istio` image are now built from [the istio-proxy repository](https://github.com/curiefense/istio-proxy/blob/1.9.2/.github/workflows/build-envoy-for-curiefense.yml)

Necessary changes were implemented:
* Use of the v3 configuration format for `envoy` instead of v3.

This PR looks big; this is due to the update of istio helm charts. To facilitate reviews, I did that in two commits:
* First, copy the charts straight from the upstream release in 1927347010f9af13ae072159bfcd2635201942ee
* Then, add curiefense to these charts in 8d787000efbc6c5358e05712a6c3d77029ff1c3c

Logs are not tested by CI, but they do work for logstash.

Signed-off-by: Xavier <xavier@reblaze.com>

PR #319, opened 2021-04-16
Move Elasticsearch config files to a contrib dir

flaper87:
A new contrib dir to host configs for different curielogger deployments is being added in this
commit. The Elasticsearch configs have been moved there and a hard-link was created in the
outputs directory.

In 1.5.0, these files will be removed along with the outputs.

related to #317


xavier-rbz:
CI currently fails, but I suspect that will get fixed by #311.

xavier-rbz (approved):
Looks good.
The commit message looks disconnected from what I see: I see 4 files being created, but no file move, and no hard links (I think git cannot represent them; symlinks are supported).

flaper87:
> Looks good.
> The commit message looks disconnected from what I see: I see 4 files being created, but no file move, and no hard links (I think git cannot represent them; symlinks are supported).

Yeah, I think git is just messing with me here. I'll get it in as-is and remove them as part of the outputs depreciation PR

PR #320, opened 2021-04-17
[docker-compose] Fix curieproxy metrics scrape

Stoakes:
Hello,

While going through the Getting started with docker-compose, I noticed that hostname of scraping target for Curiefense proxy was invalid, so here is a PR addressing it.

I also updated path configuration to match Envoy Prometheus URL, and removed unneeded `format` parameter.


flaper87 (approved):
Thanks, @Stoakes

jdorfman:
Thanks for the contribution @Stoakes! We would like to send you some swag. Email jdorfman@curiefense.io if you want to claim it. =)

PR #321, opened 2021-04-17
[istio] bump 1.9.2 -> 1.9.3

xavier-rbz:
Update istio to 1.9.3:

* update istio charts
* update curieproxy-istio docker image (`envoy` binary is built in https://github.com/curiefense/istio-proxy/commit/1fc81107d58c138eba288f3d8b567bbca4b6423f)

Fix issues:
* Increase memory limit for ingressgateway to 2Gi instead of 1Gi (value that was used before #311. When running latency tests, I see a max RAM usage of <1.5Gi)
* Fix the latency tests in `deploy-gke.sh`


phil777 (approved):
No obvious mistake.

PR #322, opened 2021-04-18
Add an stdout output

flaper87:
The original strategy for the `stdout` logger was to simply add
`os.Stdout` to the outputs enabled. This worked well for the
non-containerazed case. When running within the container, os.Stdout was
not properly flushing the buffer, causing the log to never be printed in
stdout.

Instead of playing too much with it and doing some container-specific
hacks, this commits adds a new stdout logger which has the benefit of
being consisted with the rest of the loggers. This commit appends a
new-line character at the end of the byte array passed to `os.Stdout` to
guarantee one log will be printed per line and to force flush of stdout
in the container.

In addition, we can now set custom settings for this logger, if needed.

This commit also moves the `initializer.go` module to the root `pkg` so
that we can move the `Config` struct to `config.go` and use `log_level`
as well as other global configs during the initialization of the
service. This preserves the behavior from before the refactor.

DeepinScreenshot_select-area_20210421055221
DeepinScreenshot_select-area_20210421055209

related to #317


xavier-rbz:
Since logging is not tested in PR and this touches logging, I'd like to run this PR with istio and ensure it still works. Will do it later...

maxim-kuderko:
It appears just to be a config issue of docker compose
https://github.com/docker/compose/issues/1549
setting PYTHONUNBUFFERED=1 in ENV should solve it.

**Not using the os.stdout is dangerous and could cause breaking changes in the future, i'll advice against it.**

flaper87:
> It appears just to be a config issue of docker compose
> [docker/compose#1549](https://github.com/docker/compose/issues/1549)
> setting PYTHONUNBUFFERED=1 in ENV should solve it.

mmh, I'm afraid that may not work and that it's only related to python apps. I've used it in the past for python containers. The problem with the output was present even without using `docker-compose`. Did you try this?

> 
> **Not using the os.stdout is dangerous and could cause breaking changes in the future, i'll advice against it.**

You seem to be very concerned about using `Println` vs `os.Stdout`. Would you mind elaborating on what breaking changes it may introduce? I was under the impression that `Println` was writing to `Stdout`. 

> As for the moving of initializers.go, well, It kinda breaks the separation of concerns there, the main pkg should not be aware of the output specific initialization, i'd suggest just adding the "Config" struct to the config.go and keeping the rest there

I tried just adding the `Config` struct to `config.go` but then that created circular import errors. I will give this another try as I may have messed something up while fighting with the buffer issue. Thanks for the suggestion.

xavier-rbz (approved):

PR #324, opened 2021-04-18
iptools' cargo tests for CI

phil777:


.github/workflows/run-curieconf-rust-tests.yml
```diff
@@ -27,6 +27,10 @@ jobs:
         working-directory: ./curiefense/iptools
         run: 'cargo fmt --all -- --check'
 
+      - name: Run cargo unit tests
+        working-directory: ./curiefense/iptools
+        run: 'cargo test'
+
```
flaper87, line 33: This is actually run in the `test` target of the Makefile, which we call in L50 already:

https://github.com/curiefense/curiefense/blob/7a366a4c47c957cfcabea91f07890dab9ccadd23/curiefense/iptools/Makefile#L7-L11

Maybe we should run these tests on every PR rather than running them just on Rust changes?

flaper87:
Closing for inactivity! @phil777 let me know if this is still relevant

bartavelle:
(I do not think it is as iptools should be merged with the main rust code)

PR #328, opened 2021-04-21
Add configs and templates for Elasticsearch 6.x

flaper87:


contrib/curielogger/elk/6_x/start_curielogger.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
xavier-rbz, line 1: This script doesn't actually start curielogger. Maybe rename it to "initialize_es6.sh" or something similar?

xavier-rbz (changes requested):
Could you also add a symlink this folder in curiefense/images/curielogger/contrib? That way, the contrib/curielogger scripts would be part of the curielogger image, and it could be used as an initContainer.

flaper87:
> Could you also add a symlink this folder in curiefense/images/curielogger/contrib? That way, the contrib/curielogger scripts would be part of the curielogger image, and it could be used as an initContainer.

Ah, yes, this is a good idea!

xavier-rbz (approved):

PR #330, opened 2021-04-22
Add feature to undo a version restoration through our git history component

### LINKED ISSUE #298: Version Control - Add an option to undo a version revert
In the system, we have two ways to revert the configuration to an older version. The first one is through the page "Version Contorl" and the second one is through the bottom "Verison History" of each local page

In both of them, we would like to add the ability to quickly undo a revert issued by the user. We can acomplish that by simply reverting to the second-from-last version

*Example*
Versions history (new to old): 4835254, 3007ee4, f5ade7c, d9f43dc
When the user revert from the current version (4835254) to the old version (f5ade7c), a new version number is created and the history would look like this (new to old): c721241, 4835254, 3007ee4, f5ade7c, d9f43dc
Reverting back to the second-from-last version (4835254) would let us go back to our state just before the revert took place

Aviv-Galmidi:
* Add undo version restoration functionality
* Add undo button which appears when there is an active restoration history (in the same page)
* Add unit tests for the new undo version restoration feature
* Change DocumentEditor.vue to hide gitLog while loading document instead of remove it from the DOM and reload it
* Change any usage of GitHistory.vue to reset the git log when switching log source (switch branch, doctype, etc.)
Closes #298 

Blocked by https://github.com/curiefense/curiefense/pull/327 - To keep version history intact, this PR (UI version 1.3.4) will be created as a draft until PR327 is merged (UI version 1.3.3)

### DIFF
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.3.2",
+  "version": "1.3.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/curiefense/ui/package-lock.json
+++ b/curiefense/ui/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "1.3.3",
+  "version": "1.3.4",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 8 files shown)


Aviv-Galmidi:
feature dropped

PR #332, opened 2021-04-22
[uiserver] add options to configure links to kibana&grafana

xavier-rbz:
The UI server displays links to Kibana & Grafana on the left page of
every page. The target of these links are retrieved from the
configuration server, from the "links" document of the "system" database.

This commit makes it easier to configure the target of these links by
setting two environment variables, GRAFANA_UI_URL and KIBANA_UI_URL.

### DIFF
--- a/curiefense/images/uiserver/init/start_nginx.sh
+++ b/curiefense/images/uiserver/init/start_nginx.sh
@@ -1,8 +1,30 @@
 #!/bin/bash
+
+CURL="curl --silent --write-out %{http_code}\n -H Content-Type:application/json --output /dev/null"
+
+configure_links () {
+	URL="http://confserver/api/v1/db/system/k/links/"
+	DATA="{\"kibana_url\":\"$KIBANA_UI_URL\",\"grafana_url\":\"$GRAFANA_UI_URL\"}"
+	if $CURL "$URL" -X PUT -H 'Accept: application/json' --data-raw "$DATA"| grep -qv 200; then
+		sleep 5
+		configure_links
+	fi
+}
+
+# Configure links to kibana & grafana
+if [ -n "$GRAFANA_UI_URL" ] && [ -n "$KIBANA_UI_URL" ]; then
+	echo "Updating links in the curieconf database..."
+	configure_links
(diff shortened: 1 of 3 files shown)


curiefense/images/uiserver/init/start_nginx.sh
```diff
@@ -1,8 +1,30 @@
...
+	URL="http://confserver/api/v1/db/system/k/links/"
+	DATA="{\"kibana_url\":\"$KIBANA_UI_URL\",\"grafana_url\":\"$GRAFANA_UI_URL\"}"
+	if $CURL "$URL" -X PUT -H 'Accept: application/json' --data-raw "$DATA"| grep -qv 200; then
+		sleep 5
+		configure_links
+	fi
+}
+
+# Configure links to kibana & grafana
+if [ -n "$GRAFANA_UI_URL" ] && [ -n "$KIBANA_UI_URL" ]; then
```
flaper87, line 15: Do you think this should be an `||`? I'm wondering what would happen if we want to configure just one of them

xavier-rbz: If we make this a `||`, then we need to bulid a document containing either 1 or 2 URLs. I wanted to keep this simple: either set both links, or none.
https://github.com/curiefense/curiefense/blob/main/curiefense/ui/src/components/SideMenu.vue

flaper87: @Aviv-Galmidi I think we should eventually support this. What do you think?

flaper87 (approved):
Sounds like something we can tackle in a follow-up PR

curiefense/images/uiserver/init/start_nginx.sh
Aviv-Galmidi: @flaper87 @xavier-rbz 
Currently in the UI client I check whether either of the URLs exist in system db -> links document (`kibaba_url` or `grafana_url`) and use either or both of them if they exist
I think we should support `or` pattern
But it doesn't seem that important, and as this is merged already we can support this in a future PR

PR #335, opened 2021-04-24
Switch to stdout output by default

flaper87:
Switches to using the stdout output by default. filebeat is now being setup for the docker-compose
case, instead of logstash

related to #317


deploy/curiefense-helm/curiefense/charts/curiefense-proxy/templates/curielogger-deployment.yaml
```diff
@@ -48,6 +75,14 @@ spec:
         {{ else }}
         image: {{ .Values.global.images.curielogger }}:{{ .Values.global.settings.docker_tag }}
         {{ end }}
+{{- if eq .Values.global.settings.curiefense_es_forwarder "filebeat" }}
+        args:
+          - /bin/curielogger > >(tee /var/log/curielogger/curielogger.log)
```
xavier-rbz, line 80: If I'm not mistaken, that will:
* duplicate logs to stdout (will be picked up by the cluster's log collection mechanism)
* create an endlessly growing file

Could we use a fifo instead? Something like this:
```
if [ ! -e /var/log/curielogger/curielogger.log ]; then mkfifo /var/log/curielogger/curielogger.log; fi; /bin/curielogger > /var/log/curielogger/curielogger.log
```

xavier-rbz (approved):
If you think the curielogger.log comment is irrelevant or is best handled in a subsequent PR, you may merge this.

deploy/curiefense-helm/curiefense/charts/curiefense-proxy/templates/curielogger-deployment.yaml
flaper87: You are correct! I knew this would do both things but I was planning to address it in a follow-up PR. Sorry for not having been explicit about this issue and, as usual: Good Catch! :) 

I will incorporate your suggestion as that should solve the issue perfectly!

flaper87: Here's the PR: https://github.com/curiefense/curiefense/pull/341

PR #337, opened 2021-04-24
[curielogger] do not initialize ES if not wanted

xavier-rbz:
This makes curielogger skip the initial Elasticsearch configuration if `initialize` is set to False.


curiefense/curielogger/pkg/outputs/elasticsearch.go
```diff
@@ -148,6 +148,10 @@ func (es *ElasticSearch) ConfigureKibana() {
 
 func (es *ElasticSearch) ConfigureEs() {
 	log.Warn("The Elasticsearch output is deprecated and will be removed in the 1.5.0 release. More on the reasoning and discussion here: https://github.com/curiefense/curiefense/issues/317")
+
+	if !es.cfg.Initialize {
+		return
+	}
```
flaper87, line 154: Oh, bummer. This seems to have been a regression from the refactor! :/

flaper87 (approved):
good catch!

PR #341, opened 2021-04-29
Use a FIFO file for curielogger's log file when using filebeat

flaper87:
Signed-off-by: Flavio Percoco <flaper87@gmail.com>

PR #342, opened 2021-05-01
[WIP][e2e] Fix e2e tests with minikube & filebeat

xavier-rbz:
This PR fixes bugs in #341 which I merged too soon. Apologies @flaper87...

* Fix logs display in GitHub Actions
* Reverts to passing logs from curielogger to filebeat as a file, since filebeat [does not support FIFOs](https://github.com/elastic/beats/blob/23c546eeb3064c705a82da3ac103ac341318345b/filebeat/input/log/harvester.go#L526). That means logs are written to a temporary file that is never removed

### DIFF
--- a/.github/workflows/e2e-minikube-filebeat.yml
+++ b/.github/workflows/e2e-minikube-filebeat.yml
@@ -65,9 +65,16 @@ jobs:
         if: ${{ failure() }}
         run: |
             kubectl -n curiefense describe pods
-            kubectl -n curiefense get logs -l app.kubernetes.io/name=curielogger
-            kubectl -n curiefense get logs -l app.kubernetes.io/name=curietasker
-            kubectl -n curiefense get logs -l app.kubernetes.io/name=uiserver
+            echo "--- Logs for curielogger ---"
+            kubectl -n curiefense logs --all-containers -l app.kubernetes.io/name=curielogger
+            echo "--- Logs for curietasker ---"
+            kubectl -n curiefense logs --all-containers -l app.kubernetes.io/name=curietasker
+            echo "--- Logs for uiserver ---"
+            kubectl -n curiefense logs --all-containers -l app.kubernetes.io/name=uiserver
+            echo "--- Logs for kibana ---"
+            kubectl -n curiefense logs --all-containers -l app.kubernetes.io/name=kibana
+            echo "--- Logs for elasticsearch ---"
+            kubectl -n curiefense logs --all-containers -l app.kubernetes.io/name=elasticsearch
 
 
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-minikube-filebeat.yml
```diff
@@ -65,9 +65,16 @@ jobs:
         if: ${{ failure() }}
         run: |
             kubectl -n curiefense describe pods
-            kubectl -n curiefense get logs -l app.kubernetes.io/name=curielogger
```
flaper87, line 68: I noticed this too late, but I'm glad we now have actual logs in CI!

deploy/curiefense-helm/curiefense/charts/curiefense-proxy/templates/curielogger-deployment.yaml
```diff
@@ -77,7 +77,7 @@ spec:
         {{ end }}
 {{- if eq .Values.global.settings.curiefense_es_forwarder "filebeat" }}
         args:
-          - if [ ! -e /var/log/curielogger/curielogger.log ]; then mkfifo /var/log/curielogger/curielogger.log; fi; /bin/curielogger > >(tee /var/log/curielogger/curielogger.log)
+          - /bin/curielogger > /var/log/curielogger/curielogger.log
```
flaper87, line 80: Wouldn't this capture `stdout` entirely? I think it would be good to also have it go to the console.

As far as the file itself goes, what do you think about adding logrotate to the curielogger container? It may help us deal with this and we could make this the default entrypoint and command for the curielogger container

flaper87 (approved):
LGTM. I added a question inline that we can solve in a follow up PR! The original command was using `tee` to guarantee that messages would go to both `stdout` and to the file.

deploy/curiefense-helm/curiefense/charts/curiefense-proxy/templates/curielogger-deployment.yaml
xavier-rbz: About the missing `tee`: that was not intentional... I'll open a PR to add it back.
Logrotate would be an option. That means having an additional "cleanup" container.
Another option (not convinced it's better than logrotate): configure curielogger to write to the local filesystem directly (it is [supported by gocloud](https://gocloud.dev/howto/blob/#local)). A cleanup container would just have to delete files that are older than 1 hour.

xavier-rbz: Opened #343

PR #347, opened 2021-05-07
Logrotate

flaper87:
Don't let the curielogger log file grow indefinitely. Instead, make sure it is rotated and eventually deleted.

Logrotate is configured with the following settings:

Max file size: 10MB
Number of files kept: 10
Rotation happens hourly


xavier-rbz (approved):
LGTM

PR #369, opened 2021-06-07
Initial syslog API support

flaper87:

### DIFF
--- a/curiefense/curielogger/cmd/grpc/main.go
+++ b/curiefense/curielogger/cmd/grpc/main.go
@@ -29,8 +29,10 @@ func main() {
 			pkg.InitOutputs,
 			pkg.NewMetrics,
 			pkg.NewLogSender,
+			newSyslogSrv,
 			newGrpcSrv,
 		),
+		fx.Invoke(syslogInit),
 		fx.Invoke(grpcInit),
 	)
 	if err := app.Start(context.Background()); err != nil {
--- a/curiefense/curielogger/cmd/grpc/syslog.go
+++ b/curiefense/curielogger/cmd/grpc/syslog.go
@@ -1,14 +1,14 @@
 package main
 
 import (
-	"strings"
 	"encoding/json"
+	"strings"
 
 	pkg "github.com/curiefense/curiefense/curielogger/pkg"
+	"github.com/curiefense/curiefense/curielogger/pkg/entities"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	syslog "gopkg.in/mcuadros/go-syslog.v2"
-	"github.com/curiefense/curiefense/curielogger/pkg/entities"
 )
 
 type syslogServer struct {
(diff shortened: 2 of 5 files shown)


flaper87 (commented):
CI failure seems unrelated, sending this through

curiefense/curielogger/cmd/grpc/syslog.go
```diff
@@ -0,0 +1,54 @@
...
+	server := syslog.NewServer()
+	//server.SetFormat(syslog.Automatic)
+	server.SetFormat(syslog.RFC3164)
+	server.SetHandler(handler)
+	log.Infof("syslog server listening on 9514")
+	server.ListenTCP("0.0.0.0:9514")
+	server.ListenUDP("0.0.0.0:9514")
+
+	go func(channel syslog.LogPartsChannel) {
+		for logParts := range srv.channel {
```
xavier-rbz, line 36: is our log format documented somewhere?

curiefense/curielogger/cmd/grpc/main.go
```diff
@@ -29,8 +29,10 @@ func main() {
 			pkg.InitOutputs,
 			pkg.NewMetrics,
 			pkg.NewLogSender,
+			newSyslogSrv,
 			newGrpcSrv,
 		),
+		fx.Invoke(syslogInit),
```
xavier-rbz, line 35: IIUC, curielogger will set up a syslog listener even in environments where that is not needed. How about adding a config knob so that it is only enabled in environments that need syslog?

curiefense/curielogger/cmd/grpc/syslog.go
```diff
@@ -0,0 +1,54 @@
...
+			var cfLog entities.CuriefenseLog
+			if !strings.HasPrefix(logParts["content"].(string), "nginx: ") {
+				continue
+			}
+			content := strings.TrimPrefix(logParts["content"].(string), "nginx: ")
+			err := json.Unmarshal([]byte(content), &cfLog)
+			if err != nil {
+			  log.Errorf("Error occured during unmarshaling. Error: %s", err.Error())
```
xavier-rbz, line 44:
```suggestion
			  log.Errorf("Error occurred during unmarshalling. Error: %s", err.Error())
```

xavier-rbz (commented):
general comments: I'd like to see test & documentation for the syslog support, or an issue to do that

curiefense/curielogger/cmd/grpc/main.go
flaper87: > IIUC, curielogger will set up a syslog listener even in environments where that is not needed. How about adding a config knob so that it is only enabled in environments that need syslog?

I didn't add it because having a config for this means we should also make the grpc/envoy server configurable too. I figured we would start with both enabled and make it configurable later.

curiefense/curielogger/cmd/grpc/syslog.go
flaper87: Sadly, no! It's in my todo list to document these structs beyond just code.

xavier-rbz (approved):

flaper87:
CI will be fixed by #421

PR #371, opened 2021-06-08
Use patched envoy to debug envoy issue

xavier-rbz:


tzuryby (approved):
looks like the right sha hash for the image.

PR #385, opened 2021-06-22
Rust curieproxy

bartavelle:
So this is the Rust PR. Unfortunately, it comes with other changes (mostly flowcontrol & tag actions).

It also comes with a new build process, with a Rust build container that runs Rust tests, but also a lot of mock lua tests (a bit like the e2e tests, but only testing the Rust component, and much faster).

It removes most unused Lua code too.


tzuryby (approved):
all good. 
from now on. rust is the default engine.

flaper87 (changes requested):
@bartavelle Can you check the various failures? I'd like to make sure CI passes before we send this through.

bartavelle:
Are the lua tests supposed to pass? I will fix the others right now.

curiefense/images/curieproxy-envoy/Dockerfile
```diff
@@ -1,3 +1,5 @@
+ARG RUSTBIN_TAG=latest
```
xavier-rbz, line 1: using a "latest" tag might lead to embedding a binary that is not well characterized. I'd rather have the binary built in this Dockerfile, or having a specific version instead of "latest"

xavier-rbz (changes requested):

xavier-rbz:
I'd rather have latency test results before merging this.

curiefense/curieproxy/rust/curiefense-lua/src/iptools/avltree.rs
```diff
@@ -0,0 +1,361 @@
...
+            }
+            2 => {
+                if side == InsertSide::Left {
+                    self.right.as_mut().unwrap().rotate_right();
+                }
+                self.rotate_left()
+            }
+            bf => {
+                self.dump(0);
+                panic!("Internal error: AVLTree balance factor={}. Should be in [-2;2]", bf);
```
xavier-rbz, line 193: let's never panic.

curiefense/curieproxy/rust/curiefense/src/config/profiling.rs
```diff
@@ -0,0 +1,333 @@
...
+        out.simplify();
+        out
+    }
+    fn intersection<N: iprange::IpNet>(elems: Vec<N>) -> IpRange<N> {
+        // this is a bit convoluter but the first element of the fold must be
+        // an element that is to be intersected, and not the empty set (as it
+        // would always return the empty set)
+        let mut i = elems.into_iter();
+        match i.next() {
+            None => panic!("invariant violated, elems is empty!"),
```
xavier-rbz, line 135: let's never panic. panic leads to a crash of envoy, which should never happen.

xavier-rbz (changes requested):

xavier-rbz (dismissed):
I see that this implement control flows. We need to have:
* specifications for this feature
* documentation for this feature
* e2e tests for this feature

curiefense/curieproxy/lua/nativeutils.lua
```diff
@@ -0,0 +1,166 @@
...
+function trim(s)
+    return (string.gsub(s, "^%s*(.-)%s*$", "%1"))
+end
+
+function string:startswith(arg)
+  return string.find(self, arg, 1, true) == 1
+end
+
+function string:endswith(arg)
+  return string.find(self, arg, #self - #arg + 1, true) == #self - #arg + 1
```
alexcohn-rbz, line 16: I believe that `string.sub(self, 1, #arg) == arg` is much faster especially if `arg` is short.

bartavelle:
> I see that this implement control flows. We need to have:
> 
>     * specifications for this feature
> 
>     * documentation for this feature
> 
>     * e2e tests for this feature

I wrote some specification/documentation in the usual document. I will add e2e tests now.

bartavelle:
> @bartavelle Can you check the various failures? I'd like to make sure CI passes before we send this through.

All tests are green!

flaper87:
@bartavelle is the plan to address the following comments from @xavier-rbz in this PR?

* https://github.com/curiefense/curiefense/pull/385#pullrequestreview-690438627
* https://github.com/curiefense/curiefense/pull/385#issuecomment-866684604

curiefense/images/curieproxy-envoy/Dockerfile
bartavelle: This is actually the default value, and it is overridden in `build-docker-images.sh`, so I believe we're good.

bartavelle:
@xavier-rbz @flaper87 I think the PR is in a much better shape now. Does it address your comments?

The main problem is that the new features do not have e2e tests, but they all have "lua tests" that run during the rust build phase.

flaper87:
> The main problem is that the new features do not have e2e tests, but they all have "lua tests" that run during the rust build phase.

This is a more promising situation. I think we can start getting this in and work on the remaining tasks on follow-up PRs.

Can you work on an e2e tests for the new feature as the next task?

I will squash this into a single commit. It's a quite big PR, which is unfortunate :)

flaper87 (approved):

PR #396, opened 2021-06-28
chore: Use --no-cache-dir flag to pip in Dockerfiles, to save space

Rajpratik71:
Using "--no-cache-dir" flag in pip install ,make sure dowloaded packages
by pip don't cached on system . This is a best practise which make sure
to fetch ftom repo instead of using local cached one . Further , in case
of Docker Containers , by restricing caching , we can reduce image size.
In term of stats , it depends upon the number of python packages
multiplied by their respective size . e.g for heavy packages with a lot
of dependencies it reduce a lot by don't caching pip packages.

Further , more detail information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/curiefense/images/confserver/Dockerfile
+++ b/curiefense/images/confserver/Dockerfile
@@ -16,6 +16,6 @@ RUN /usr/bin/apt-get update && \
 COPY bootstrap /bootstrap
 COPY init /init
 COPY curieconf-utils /curieconf-utils
-RUN cd /curieconf-utils ; pip3 install .
+RUN cd /curieconf-utils ; pip3 install --no-cache-dir .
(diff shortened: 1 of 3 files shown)


jdorfman:
Thanks @Rajpratik71! We will review it asap.

flaper87 (approved):
Thanks a lot, @Rajpratik71 

LGTM, let's wait for CI now :)

jdorfman:
@flaper87 is the e2e failed test on our end or something we need @Rajpratik71 to look into?

flaper87 (approved):
failure seems unrelated

Issue #370, opened 2021-06-08
Nginx integration


bartavelle:
**Is your feature request related to a problem? Please describe.**
As the Curieproxy filtering feature is now implemented as a native library, it can be decoupled from Envoy.
Existing nginx (openresty) users could "upgrade" to Curieproxy.

**Describe the solution you'd like**
The solution is kind of done already. Just like the Envoy solution, it is based on a layer of Lua that calls the native library, and interprets its output.

Logging is performed by putting the JSON-encoded data in an nginx variable, that can then be exposed (using syslog, local files, pipes, etc.) by nginx.

Work done:

  * Lua code that calls the native code ;
  * Lua code that handles logging ;
  * Docker container that integrates with the current infrastructure

To be done:

 * Integration with Syslog logging (see https://github.com/curiefense/curiefense/pull/369)
 * Integration with E2E testing

**Additional context**
The biggest problem here is how to prepare the PR/merge ...

PR #400, opened 2021-07-01
Add an nginx-ingress container

flaper87:
This container is build from the curieproxy-nginx container, openresty container, and kubernetes-ingress container.
It allows to use the nginx-ingress controller with openresty, which is a build of nginx with support for external
lua code execution.

### DIFF
--- a/curiefense/images/curiefense-nginx-ingress/Dockerfile
+++ b/curiefense/images/curiefense-nginx-ingress/Dockerfile
@@ -1,0 +1,65 @@
+ARG RUSTBIN_TAG=latest
+FROM curiefense/curiefense-rustbuild-focal:${RUSTBIN_TAG} AS rustbin
+FROM docker.io/nginx/nginx-ingress:1.11.3 as nginx-ingress
+FROM docker.io/openresty/openresty:focal as openresty
+
+USER root
+RUN groupadd --system --gid 101 nginx \
+    && useradd --system --gid nginx --no-create-home --home-dir /nonexistent --comment "nginx user" --shell /bin/false --uid 101 nginx 
+
+RUN apt-get update \
+    && apt-get install -y \
+        libcap2-bin \
+        libhyperscan5 \
+        rsyslog \
+	python3-pip \
+        python3-setuptools \
+        python3-wheel \
+        python3-magic \
+	python3-xattr \
+        python3-netifaces \
+    && rm -rf /var/lib/apt/lists/*
+
+# bucket credentials will be provided by docker-compose or k8s secrets, if needed
+RUN ln -s /run/secrets/s3cfg/s3cfg /root/.s3cfg
+
+COPY curieconf-utils /curieconf-utils
+RUN cd /curieconf-utils ; pip3 install .
+
+COPY curieconf-client /curieconf-client
+RUN cd /curieconf-client ; pip3 install .
+
+COPY init /curiesync
+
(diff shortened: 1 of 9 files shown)


curiefense/images/curiefense-nginx-ingress/rsyslog.conf
```diff
@@ -0,0 +1,16 @@
...
+$FileGroup adm
+$FileCreateMode 0640
+$DirCreateMode 0755
+$Umask 0022
+$PrivDropToUser syslog
+$PrivDropToGroup syslog
+$WorkDirectory /var/spool/rsyslog
+
+*.* /var/log/syslog
+action(type="omfwd" Target="curielogger" Port="9514" Protocol="tcp")
```
flaper87, line 15: This will have to change, since in a k8s environment we need to use curielogger's internal svc address.

bartavelle:
Why not replace the curiefense-nginx image?

flaper87:
> Why not replace the curiefense-nginx image?

I guess we could have a single image for both. I wouldn't replace the `curiefense-nginx` image since that's useful when we are not running in kubernets. The `ngnix-ingress` is Kubernetes specific

curiefense/images/curiefense-nginx-ingress/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+	&& chown -R nginx:0 /etc/nginx /etc/nginx/secrets /var/cache/nginx /var/lib/nginx /nginx* \
+        && chown -R nginx:0 /var/lib/openresty /var/run/openresty /var/cache/openresty /usr/local/openresty \
+	&& setcap 'cap_net_bind_service=+ep' /usr/local/openresty/nginx/sbin/nginx \
+	&& setcap -v 'cap_net_bind_service=+ep' /usr/local/openresty/nginx/sbin/nginx  \
+	&& rm -f /etc/nginx/conf.d/*
+
+RUN sed -i "/^http {/a log_format curiefenselog escape=none '$request_map';" /nginx.tmpl \
+        && sed -i "/^http {/a lua_package_path '/lua/?.lua;;';" /nginx.tmpl
+
+RUN openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 -subj "/C=fr/O=curiefense/CN=testsystem" -keyout /etc/ssl/certificate.key -out /etc/ssl/certificate.crt
```
xavier-rbz, line 49: what are this key and cert used for?

xavier-rbz (commented):
I'd love to see tests (is that covered by #370?) & documentation for this container in the future.
I'm fine with it being merged if an issue for test&doc is open!

curiefense/images/curiefense-nginx-ingress/Dockerfile
flaper87: > what are this key and cert used for?

I think we can do away with them, I took it from the existing nginx image.

flaper87:
> I'd love to see tests (is that covered by #370?) & documentation for this container in the future.
> I'm fine with it being merged if an issue for test&doc is open!

Not coming in 370 but this does depend on that. I'm working on the tests for it, the same way as we are testing istio on minilube. :D

xavier-rbz (approved):

flaper87:
CI will be fixed by https://github.com/curiefense/curiefense/pull/421

PR #421, opened 2021-07-08
update minikube to fix CI

xavier-rbz:
Prevents an error that looks very similar to
https://github.com/kubernetes/minikube/pull/11419


flaper87 (approved):
Thanks! :pray:

PR #423, opened 2021-07-08
UI error handling

denis-reblaze:


curiefense/ui/src/assets/RequestsUtils.ts
```diff
@@ -56,15 +56,29 @@ const processRequest = (methodName: MethodNames, apiUrl: string, data: any, conf
...
+  url: string,
+  data?: any,
+  config?: AxiosRequestConfig,
+  successMessage?: string,
+  failureMessage?: string,
+  undoFunction?: () => any,
+  onFail?: Function,
+}
+
+const sendRequest = (requestParams: IRequestParams) => {
```
Aviv-Galmidi, line 78: What is the reason behind changing this function to receive an object instead of params?

curiefense/ui/src/views/DBEditor.vue
```diff
@@ -285,7 +290,7 @@ export default Vue.extend({
       keyNameInput: '',
       defaultKeyName: 'publishinfo',
 
-      selectedDatabaseData: null,
+      selectedDatabaseData: {} as {[key: string]: string},
```
Aviv-Galmidi, line 293: The value type is not a string but a json, you can use the type GenericObject to represent this

curiefense/ui/src/views/DBEditor.vue
```diff
@@ -358,7 +362,8 @@ export default Vue.extend({
       this.setLoadingDocStatus(true)
       this.selectedDatabase = database
       this.databaseNameInput = this.selectedDatabase
-      this.selectedDatabaseData = (await RequestsUtils.sendRequest('GET', `db/${this.selectedDatabase}/`)).data
+      const response = await RequestsUtils.sendRequest({methodName: 'GET', url: `db/${this.selectedDatabase}/`})
+      this.selectedDatabaseData = response ? response.data || {} : {}
```
Aviv-Galmidi, line 366: We can use elvis opreator (optional chaining) here  `this.selectedDatabaseData = response?.data || {}`

curiefense/ui/src/views/DBEditor.vue
```diff
@@ -87,6 +89,7 @@
                   <button class="button is-small fork-key-button"
                           :class="{'is-loading': isForkKeyLoading}"
                           @click="forkKey"
+                          :disabled="!selectedDatabase"
```
Aviv-Galmidi, line 92: Perhaps this should check for both selectedDatabase and selectedKey

curiefense/ui/src/views/DocumentEditor.vue
```diff
@@ -483,8 +490,9 @@ export default Vue.extend({
     async switchDocID() {
       this.setLoadingDocStatus(true)
       this.loadGitLog()
+      const docName = (this.selectedDoc as BasicDocument)?.name
       Utils.toast(
-          `Switched to document "${(this.selectedDoc as BasicDocument).name}" with ID "${this.selectedDocID}".`,
+          `Switched to document ${docName ? `"${docName}" ` : ''}with ID "${this.selectedDocID}".`,
```
Aviv-Galmidi, line 495: It would be better to not toast if the document name is empty, or toast with something like 'Switched to unknown document'

curiefense/ui/src/views/VersionControl.vue
```diff
@@ -146,6 +146,7 @@
         <hr/>
         <git-history :gitLog="gitLog"
                      :apiPath="gitAPIPath"
+                     :loading="loadingGitlog"
```
Aviv-Galmidi, line 149: There are other locations which use gitHistory component which will need this change as well

Aviv-Galmidi (changes requested):

curiefense/ui/src/assets/RequestsUtils.ts
denis-reblaze: I had to add an argument onFail which is a callback executing when axios request fails. But the function already has 7 arguments, some of which are optional, so through the code we have such examples as
`RequestsUtils.sendRequest('GET', url, null, null, something, null, somethingElse)`

So the option to simply add an argument would be the worst decision as it would require from developer to write smth like
`RequestsUtils.sendRequest('GET', url, null, null, null, null, null, onFail)`

Anyway, regardless the current PR, this function definitely must be refactored.

curiefense/ui/src/views/DBEditor.vue
denis-reblaze: If selectedDatabase is falsy then selectedKey is definitely falsy, no reason to check it

curiefense/ui/src/assets/RequestsUtils.ts
Aviv-Galmidi: What refactor would you like to see in the funciton? Let's dicuss this

denis-reblaze: the same as described above.
in addition, but not that necessarily, some lines could be simplified

Aviv-Galmidi (approved):

PR #425, opened 2021-07-10
Allow to running rsyslog as non-privileged user

flaper87:
Remove bind to 514, and uses /var/lib/syslog/log as the socket
path instead of /dev/log

### DIFF
--- a/curiefense/images/curiefense-nginx-ingress/rsyslog.conf
+++ b/curiefense/images/curiefense-nginx-ingress/rsyslog.conf
@@ -1,14 +1,9 @@
-module(load="imuxsock") # provides support for local system logging
-module(load="imudp")
-input(type="imudp" port="514")
+module(load="imuxsock" SysSock.Name="/var/lib/syslog/log")
 
-$FileOwner syslog
 $FileGroup adm
 $FileCreateMode 0640
 $DirCreateMode 0755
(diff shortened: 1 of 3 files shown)


xavier-rbz (approved):

curiefense/images/curiefense-nginx-ingress/rsyslog.conf
```diff
@@ -1,16 +1,13 @@
...
-$FileOwner syslog
 $FileGroup adm
 $FileCreateMode 0640
 $DirCreateMode 0755
 $Umask 0022
-$PrivDropToUser syslog
-$PrivDropToGroup syslog
 $WorkDirectory /var/spool/rsyslog
 
 *.* /var/log/syslog
```
bartavelle, line 11: This will write in the container, and should probably go once everything is working.

bartavelle (approved):
Looks good to me.

curiefense/images/curiefense-nginx-ingress/rsyslog.conf
flaper87: > This will write in the container, and should probably go once everything is working.

Indeed! keeping it around for now for debuggability purposes but will remove soon :)

tzuryby (approved):

PR #442, opened 2021-07-12
Properly logs attrs

bartavelle:
Add missing fields, such as:

 * authority
 * method
 * content of mextra (items passed as meta that are not used by
   the filtering module)

### DIFF
--- a/curiefense/curieproxy/rust/curiefense/src/utils.rs
+++ b/curiefense/curieproxy/rust/curiefense/src/utils.rs
@@ -197,24 +197,37 @@ pub struct RequestInfo {
 
 impl RequestInfo {
     pub fn into_json(self, tags: Tags) -> serde_json::Value {
-        // TODO: geo
         let ipnum: Option<String> = self.rinfo.geoip.ip.as_ref().map(|i| match i {
             IpAddr::V4(a) => u32::from_be_bytes(a.octets()).to_string(),
             IpAddr::V6(a) => u128::from_be_bytes(a.octets()).to_string(),
         });
+        let geo = self.rinfo.geoip.to_json();
+        let mut attrs: HashMap<String, Option<String>> = [
+            ("uri", self.rinfo.qinfo.uri),
+            ("path", Some(self.rinfo.qinfo.qpath)),
+            ("query", Some(self.rinfo.qinfo.query)),
+            ("ip", Some(self.rinfo.geoip.ipstr)),
+            ("ipnum", ipnum),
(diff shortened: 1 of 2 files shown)


flaper87 (dismissed):

curiefense/curieproxy/rust/curiefense/src/utils.rs
```diff
@@ -197,24 +197,37 @@ pub struct RequestInfo {
...
 impl RequestInfo {
     pub fn into_json(self, tags: Tags) -> serde_json::Value {
-        // TODO: geo
         let ipnum: Option<String> = self.rinfo.geoip.ip.as_ref().map(|i| match i {
             IpAddr::V4(a) => u32::from_be_bytes(a.octets()).to_string(),
             IpAddr::V6(a) => u128::from_be_bytes(a.octets()).to_string(),
         });
+        let geo = self.rinfo.geoip.to_json();
+        let mut attrs: HashMap<String, Option<String>> = [
+            ("uri", self.rinfo.qinfo.uri),
```
tzuryby, line 206: Why don't you simply loop inside all entries within attrs?
Basically, every ":foo" should be there (in attr)

tzuryby (approved):
as suggested, looping through all attributes and metadata storing them for the log would be future compatible and avoid what is potentialy an error prone.

curiefense/curieproxy/rust/curiefense/src/utils.rs
bartavelle: I am not sure what you mean, as `attrs` is being built here, we can't loop into it.

bartavelle: Ah I got it. Unrecognized arguments are stored in the mextra map, which is added to attrs, so this should be future proof.

curiefense/curieproxy/rust/curiefense/src/utils.rs
```diff
@@ -197,24 +197,37 @@ pub struct RequestInfo {
...
+            ("query", Some(self.rinfo.qinfo.query)),
+            ("ip", Some(self.rinfo.geoip.ipstr)),
+            ("ipnum", ipnum),
+            ("authority", Some(self.rinfo.host)),
+            ("method", Some(self.rinfo.meta.method)),
+        ]
+        .iter()
+        .map(|(k, v)| (k.to_string(), v.clone()))
+        .collect();
+        attrs.extend(
```
bartavelle, line 217: It is added here

tzuryby (approved):

PR #470, opened 2021-07-13
Default blocking status should be 403

bartavelle:
And not 503!


flaper87 (approved):
🍕

Issue #455, opened 2021-07-13
Tags with underscore are not working


yuvalbaryochay:
**Describe the bug**
I created a tag rule with the tag `test_01`, and matched allowed policy while traffic was blocked.

After removing the underscore, all worked.

either support underscores or prevent the UI and the API from using them.

Aviv-Galmidi:
Re-opening so we could run QA tests to make sure it is fixed
@yanushg tagging for notice

semharel:
It is Working

PR #473, opened 2021-07-14
Tags underscore

denis-reblaze:
Disallow non alphanum and ':' symbols in tags
tags are all lowercase too


curiefense/ui/src/assets/Utils.ts
```diff
@@ -117,11 +117,14 @@ const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () =
   return element
 }
 
+const filterTag = (tag: string) => tag.replace(/[^\w: ]|_/g, '-')
```
Aviv-Galmidi, line 120: As this is something which we will only use in relations to tags I think we should move this to TagsAutoComplete component, I don't think we'll be using this function anywhere else

curiefense/ui/src/components/AutocompleteInput.vue
```diff
@@ -110,8 +111,9 @@ export default (Vue as VueConstructor<Vue & {
   },
 
   data() {
+    const {filterFunction, initialValue} = this
     return {
-      autocompleteValue: this.initialValue,
+      autocompleteValue: filterFunction ? filterFunction( initialValue ) : initialValue,
```
Aviv-Galmidi, line 116: Please make sure we also add this filter function on the watch of `initialValue`

Aviv-Galmidi (changes requested):

curiefense/ui/src/components/AutocompleteInput.vue
```diff
@@ -85,12 +85,14 @@ export default (Vue as VueConstructor<Vue & {
...
+    filterFunction: Function,
   },
 
   watch: {
-    initialValue: function(newVal) {
-      if (this.autocompleteValue !== newVal) {
-        this.autocompleteValue = newVal
+    initialValue( newVal ) {
+      const newValFiltered = this.filterFunction ? this.filterFunction( newVal ) : newVal
+      if ( this.autocompleteValue !== newVal ) {
```
Aviv-Galmidi, line 94: Should be `this.autocompleteValue !== newValFiltered` as it might be the same as new value but not the filtered new value
For instance if the filter is `(val) => val * 2` and we pass 2 as the initial value and then update it to 4, autocompleteValue would be 4 which would cause it to not update

denis-reblaze: No, this will make the value filtered immediately on keypress while we want filter it on submit (on pressing space or etc.)

Aviv-Galmidi:
Adding github hooks: This PR closes https://github.com/curiefense/curiefense/issues/455

curiefense/ui/src/components/AutocompleteInput.vue
Aviv-Galmidi: But right now it can cause an issue as I described above
> if the filter is (val) => val * 2 and we pass 2 as the initial value and then update it to 4, autocompleteValue would be 4 which would cause it to not update
We'll need to think of a way to update the value on submit but also on initialValue changes

Aviv-Galmidi: Can't think of a real-world scenario where this would apply in, as replacing strings would always be different than the newVal if newVal is invalid and thus end up with the value being filtered. resolving the convo

Aviv-Galmidi (approved):

PR #481, opened 2021-07-15
Wafonly inspect

bartavelle:
This is the "WAF only" API feature, a function that only exposes the WAF filter, for integration with other tools.

### DIFF
--- a/curiefense/curieproxy/rust/NOTES.md
+++ b/curiefense/curieproxy/rust/NOTES.md
@@ -2,8 +2,7 @@
 
 An initial implementation of all filtering components had been written. The following is currently handled in Lua:
 
- * getting the information required by the filtering engine:
-    - 
+ * getting the information required by the filtering engine
  * request replies
  * logging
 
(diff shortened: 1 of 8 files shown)


curiefense/curieproxy/rust/NOTES.md
```diff
@@ -39,6 +38,46 @@ It will perform all the curieproxy checks, and return a pair, with:
...
+     --   * authority : optionally, the HTTP2 authority field
+     local response
+-    response, err = curiefense.inspect_request(
+-        meta, headers, body_content, ip_str, grasshopper
++    response, err = curiefense.inspect_waf(
++        meta, headers, body_content, ip_str, "__default__"
+     )
+ 
+     if err then
+```
```
flaper87, line 79: Is this a temporary requirement? It may not be realistic to require patching nginx session to use this feature

bartavelle: This is not supposed to be used with curiefense, but with another product. This is just a demonstration on how the feature could be used.

curiefense/curieproxy/rust/curiefense-lua/src/lib.rs
```diff
@@ -16,6 +16,84 @@ use curiefense::interface::{Decision, Grasshopper};
...
+        HashMap<String, String>, // meta
+        HashMap<String, String>, // headers
+        Option<LuaString>,       // maybe body
+        String,                  // ip
+        String,                  // waf_id
+    ),
+) -> LuaResult<(String, Option<String>)> {
+    let (meta, headers, lua_body, str_ip, waf_id) = args;
+
+    // TODO: solve the lifetime issue for the &[u8] to reduce duplication
```
Aviv-Galmidi, line 47: Did you include the ToDo on purpose or was it something you wanted to do before the PR?

bartavelle: This is actually something I solved! Pushing a new PR without the comment.

Aviv-Galmidi (approved):

PR #482, opened 2021-07-15
Implement automatic base64 decoding

bartavelle:
Works on body, args, cookies, headers. This automatically try to decode using "vanilla" base64 anything that will be a well formed utf-8 string once decoded, and allows matching it.

### DIFF
--- a/curiefense/curieproxy/rust/curiefense/Cargo.toml
+++ b/curiefense/curieproxy/rust/curiefense/Cargo.toml
@@ -9,6 +9,7 @@ crate-type = ["lib"]
 bench = false
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


flaper87:
> Works on body, args, cookies, headers. This automatically try to decode using "vanilla" base64 anything that will be a well formed utf-8 string once decoded, and allows matching it.

Is this a feature that existed in the old lua code that you are implementing now? Or is it a new feature?

It would be nice to add more information about the details and goals of the feature in the PR description. If anything, it will help for future investigation :)

flaper87 (approved):

Aviv-Galmidi (approved):

PR #486, opened 2021-07-18
response.headers should be a flattened/object, nt keyword

flaper87:


bartavelle (approved):

tzuryby (commented):
LGTM

Issue #479, opened 2021-07-14
Header action in rate limit\tag rules\ flow control should contain key:value spaces 


yanushg:
**Describe the bug**
Header action in rate limit\tag rules\ flow control should contain key:value spaces 
**To Reproduce**
1. Go to Rate limit and change the action to header 

**Expected behavior**
Header action should have key:value sections - we have only key 
**Screenshots**
Screenshot of how it should look - 
<img width="443" alt="Screen Shot 2021-07-14 at 16 29 00" src="https://user-images.githubusercontent.com/48092909/125630822-a1cc96e1-8f43-48b1-8d4e-0e68add51827.png">

yanushg:
@Aviv-Galmidi @bartavelle I think this is a combination of FE and BE effort

denis-reblaze:
As agreed with @yanushg, not only `Request Header` should contain the pair (Header Name, Header Value) but also `Response`. Also as part of the current bugfix action response fields validation was implemented. The validation rules are:

1) status: not empty and matching `^[1-5][0-9][0-9]$`
2) ttl: not empty, numeric, and < 86400
3) location: not empty and matching ^https?://.+$
4) header name, header value: not empty and matching ^[0-9a-zA-Z-_]+$

![screenshot-localhost_8080-2021 07 20-15_48_27](https://user-images.githubusercontent.com/86227793/126326922-b00e9151-0a84-47a0-9a3a-793391569147.png)

Aviv-Galmidi:
Changing milestone 1.4.0 -> 1.4.1 which is set as the next release as 1.4.0 has been released last night

PR #492, opened 2021-07-19
Add key field to header action

denis-reblaze:
added header value field and all fields validations in response actions component


curiefense/ui/src/assets/Utils.ts
```diff
@@ -93,6 +93,9 @@ const toast = (message: string | HTMLElement, type: ToastType, undoFunction?: ()
   })
 }
 
+// Closes all toast messages if there're error toasts
+const closeAllToasts = () => document.querySelector( '.notification.is-danger' )?.closest( 'div' )?.remove()
```
Aviv-Galmidi, line 97: This does not close all toasts but just toasts with `is-danger` class

Aviv-Galmidi: Why do we need this function? we shouldn't close the toasts for the user unpredictability like that

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -18,19 +18,18 @@
       12
       12
       -->
-      <div v-if="labelDisplayedInline"
-           class="column is-2">
+      <div v-if="labelDisplayedInline" class="column is-2">
```
Aviv-Galmidi, line 21: each attribute should be in a new line according to vue.js standards
https://vuejs.org/v2/style-guide/#Multi-attribute-elements-strongly-recommended

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -43,41 +42,39 @@
...
-              @change="emitActionUpdate"
+              @blur="fieldChange( 'status' )"
+              @input="fieldChange( 'status' )"
               title="Status code"
-              placeholder="Status code">
-          <span v-if="localAction && localAction.type === 'ban'"
-                class="suffix seconds-suffix">
+              placeholder="Status code"
+              :ref="localAction.type"
+          />
```
Aviv-Galmidi, line 63: Please refrain from starting a new line for the html closing tag, I'll add this rule to eslint rules set as well

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -210,28 +264,108 @@ export default Vue.extend({
...
+              this.fieldErrors.push( `headers-${field}` )
+              message.push(
+                value === '' ? `Header ${field} shouldn't be empty` : `Header ${field} doesn't match ^[0-9a-zA-Z-_]+$`,
+              )
+            }
+          })
+        }
+      }
+      if ( message.length ) {
+        Utils.toast( message.join( '<br />' ), 'is-danger' )
```
Aviv-Galmidi, line 332: The toasting mechanism is not intended for temporary form validation erros display. Please refrain using it this way.

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -210,28 +264,108 @@ export default Vue.extend({
...
+    fieldChange( field: string ) {
+      if ( ['headers-name', 'headers-value'].includes( field )) {
+        this.localAction.params.headers = {}
+        this.localAction.params.headers[this.tempHeader.name as string] = this.tempHeader.value
+      }
+      this.emitActionUpdate()
+      this.validate()
+    },
+
+    validate() {
```
Aviv-Galmidi, line 283: Please separate this function into multiple validation functions (one per field type) and use `Utils.validateInput` function

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -210,28 +264,108 @@ export default Vue.extend({
...
+      const index = this.fieldErrors.indexOf( val )
+      if ( index > -1 ) {
+        this.fieldErrors.splice( index, 1 )
+      }
+    },
   },
   watch: {
     action: {
-      handler: function(value) {
+      handler( value ) {
```
Aviv-Galmidi, line 346: I'm not sure if the binding for `this` would work correctly here, what is the reasoning behind applying this change?

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -241,5 +375,9 @@ export default Vue.extend({
   padding-top: 0;
 }
 
+.help.is-danger {
```
Aviv-Galmidi, line 378: What is this used for?

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -171,11 +208,14 @@ export default Vue.extend({
...
       }),
+      tempHeader: {name: '', value: ''},
+      fieldErrors: [],
+      isValid: true,
     }
   },
   computed: {
     localAction(): ResponseActionType {
-      return _.cloneDeep(this.action)
+      return _.cloneDeep(this.action) || {type: 'default'}
```
Aviv-Galmidi, line 218: This is not a good practice!
`localAction` is a clone of the prop `action`, if you add this code then we would have discrepancy as the two variables would not match each other and we wouldn't know about it

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -104,25 +101,64 @@
...
     </div>
-    <div class="content" v-if="localAction && localAction.type === 'ban' && localAction.params.action">
-      <response-action :action.sync="localAction.params.action"
-                       :label-separated-line="labelSeparatedLine"
-                       :is-single-input-column="isSingleInputColumn"
-                       :ignore="['ban']"
-                       @update:action="emitActionUpdate"
-                       label="Ban action"/>
+    <div
+      v-if="['request_header', 'response'].includes( localAction.type )"
```
Aviv-Galmidi, line 119: Why did you add the header to 'response' action as well? Please make sure it is needed and if so add a comment to the ticket explaining the difference in the implementation

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -104,25 +101,64 @@
...
+      class="columns"
+    >
+      <div class="column" :class="{
+        'is-6': !labelDisplayedInline || isSingleInputColumn,
+        'is-5 is-offset-2': labelDisplayedInline && !isSingleInputColumn,
+        }">
+        <input
+          v-model="tempHeader.name"
+          class="input is-small action-headers"
+          :class="{'is-danger': fieldErrors.includes ( 'headers-name' )}"
```
Aviv-Galmidi, line 129: We can set the validation of a specific field in a computed property instead of maintaining a new array for errors, each field would have it's own computed property which checks validation, and would display an error / emit invalidity as needed

Aviv-Galmidi (changes requested):
There are many undocumented changes you've done here that do not comply with the opened ticket, please make sure that everything extra you're doing is documented well

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: Please provide the desirable mechanism of showing errors.

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: Why on Earth? The standards to which you referred describes this style as the correct one. I hate looking for the closing tag through the code, it increases the development time.
https://vuejs.org/v2/style-guide/#Multi-attribute-elements-strongly-recommended

curiefense/ui/src/assets/Utils.ts
denis-reblaze: > This does not close all toasts but just toasts with `is-danger` class

This does, as all the toasts are in the same parent div.

denis-reblaze: > Why do we need this function? we shouldn't close the toasts for the user unpredictability like that

I used toasts for displaying errors, sometimes they need to be cleared. I'm removing this as you going to provide another method of displaying errors.

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: fixed

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: the reasoning is the readability increasing, handler( value ) {} and handler: function(value) {} are synonyms, the different notation is handler: (value) => {} which will cause binding issues

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: fixed

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: https://github.com/curiefense/curiefense/issues/479#issuecomment-883366102

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: What's the problem with maintaining of errors array? The logic is clear enough, the code is well-readable. On the other hand, keeping several different variables (which are technically computed properties) decreases the readability making errors dispersed through the code instead of keeping them in a single place.

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: Having a default value for the main local variable is the best practice: the whole component doesn't make sense without  the `action` prop, so it's better to have a default value instead of never ending verifications like
`v-if="localAction && (localAction.type === 'response')"`
The only place where would we have discrepancy is `(!_.isEqual(this.localAction, this.action))` - I agree, this should be complemented as `(this.action && !_.isEqual(this.localAction, this.action))`

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: You're requesting to completely rebuild the code while the idea was just to copy and adapt the existing one. If I'll do the rebuild then let's extend the task's time.

flaper87:
@denis-reblaze would you mind cleaning up the PR? It looks like there are many merges in the git history that should not be there :thinking:

curiefense/ui/src/components/ResponseAction.vue
Aviv-Galmidi: The standards which I referred to do not describe closing the html tag in a new line, they do so but they do not set it as "best practices" as it is a personal preference and not a standard
Their eslint rule `vue/html-closing-bracket-newline` is set to enforce this style by default, but as it's a personal linting preference it can be changed
I personally do not believe in this style as in my honest opinion it makes the code harder to read and understand

curiefense/ui/src/components/ResponseAction.vue
Aviv-Galmidi: Red border around the invalid field, red text next to (or under) the input field
We don't have a set styleguide for this yet and it will be created soon. so far now just a text message next to the invalid field would work

curiefense/ui/src/components/ResponseAction.vue
Aviv-Galmidi: If the validation do not comply with the time we have at the moment for the release, then let's add them properly when we do have time in 1.5.0. I do not believe we should add this code as it is right now as it is overly complicated and should be refactored

curiefense/ui/src/components/ResponseAction.vue
Aviv-Galmidi: Even if the entire component doesn't make sense without a given `action` prop, adding this code would break the symmetry between `action` and `localAction`
For the user it would seem everything is okay, but as they try to save the `action` would be empty and thus cause an unseen error

Note that if there is no `action` in prop we normallize it below in the watcher and emit `{type: 'default'}` so it would trinkle back down to this component

curiefense/ui/src/components/ResponseAction.vue
Aviv-Galmidi: I do not see the added value of the errors array

Each field has its own validators, it's own error message, etc. Merging them into a single array and function causes said function to be very large and include a lot of data, which in turn causes it to be harder to maintain and read. Each function should have a single responsibility

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: No it wouldn't

screenshot-localhost_8080-2021 07 22-11_27_37
screenshot-localhost_8080-2021 07 22-11_30_21
screenshot-localhost_8080-2021 07 22-11_30_53

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: ok

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: Ok, since your personal preferences are not a standard, you can't decide to set them as a rule without a discussion. I completely disagree and for me this style is much harder to read.

Aviv-Galmidi: We currently have a standard in Curiefense - all tags are closed without a new line. Even if this standard is not currently enforced by linters
If you want to change the standard we can have a discussion about it, but I'm not approving code which does not comply with these standards

denis-reblaze:
> @denis-reblaze would you mind cleaning up the PR? It looks like there are many merges in the git history that should not be there 🤔

fixed, hopefully...

flaper87 (commented):
LGTM, I'll let @Aviv-Galmidi do the final approval

curiefense/ui/src/components/ResponseAction.vue
```diff
@@ -104,25 +95,58 @@
...
       </template>
     </div>
-    <div class="content" v-if="localAction && localAction.type === 'ban' && localAction.params.action">
-      <response-action :action.sync="localAction.params.action"
-                       :label-separated-line="labelSeparatedLine"
-                       :is-single-input-column="isSingleInputColumn"
-                       :ignore="['ban']"
-                       @update:action="emitActionUpdate"
-                       label="Ban action"/>
+    <div
```
Aviv-Galmidi, line 110: Let's use the existing columns instead of creating new ones if possible

curiefense/ui/src/types/index.d.ts
```diff
@@ -47,7 +47,7 @@ declare module CuriefenseClient {
     params?: {
       status?: string
       ttl?: string
-      headers?: string
+      headers?: GenericObject
```
Aviv-Galmidi, line 50: Please make sure that this is the format the backend is expecting, as it's a structure change and according to @bartavelle the only change needed is in the Client

denis-reblaze: the backend expects headers to be `HashMap<String, String>`

Aviv-Galmidi:
@denis-reblaze 
I've downloaded this branch and tested it locally, and the headers are not being added to the responses as they should be
According to the rust code found [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/interface.rs#L233) and [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/config/raw.rs#L185) the correct format for headers is `Option<String>`. I'm not sure where you've gotten the `HashMap<String, String>` from, if you could direct me to the matching line that would be appreciated.

In addition, my earlier comment
> Let's use the existing columns instead of creating new ones if possible
was referring to the bulma `columns` class, not code indentations. We already have a `columns` class separating the entire ResponseAction component into columns and lines, there's no need for another wrapper

denis-reblaze:
> @denis-reblaze
> I've downloaded this branch and tested it locally, and the headers are not being added to the responses as they should be
> According to the rust code found [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/interface.rs#L233) and [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/config/raw.rs#L185) the correct format for headers is `Option<String>`. I'm not sure where you've gotten the `HashMap<String, String>` from, if you could direct me to the matching line that would be appreciated.
> 
> In addition, my earlier comment
> 
> > Let's use the existing columns instead of creating new ones if possible
> 
> was referring to the bulma `columns` class, not code indentations. We already have a `columns` class separating the entire ResponseAction component into columns and lines, there's no need for another wrapper
> 
> edit: formatting

https://github.com/curiefense/curiefense/blob/ef7ee2916ccf203906472ee639e255f2af70bd30/curiefense/curieproxy/rust/curiefense/src/interface.rs#L120-L129

Aviv-Galmidi:
> > @denis-reblaze
> > I've downloaded this branch and tested it locally, and the headers are not being added to the responses as they should be
> > According to the rust code found [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/interface.rs#L233) and [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/config/raw.rs#L185) the correct format for headers is `Option<String>`. I'm not sure where you've gotten the `HashMap<String, String>` from, if you could direct me to the matching line that would be appreciated.
> > In addition, my earlier comment
> > > Let's use the existing columns instead of creating new ones if possible
> > 
> > 
> > was referring to the bulma `columns` class, not code indentations. We already have a `columns` class separating the entire ResponseAction component into columns and lines, there's no need for another wrapper
> > edit: formatting
> 
> https://github.com/curiefense/curiefense/blob/ef7ee2916ccf203906472ee639e255f2af70bd30/curiefense/curieproxy/rust/curiefense/src/interface.rs#L120-L129

That's weird. I'm not sure why there is a discrepancy between the two `Action` types... It looks like the lines I linked were changed from `HashMap<string, string>` to `string`, perhaps there is additional change that is needed to be done in the rust code...

Anyway, did you test this feature? Does it work for you? Because as I've written in the previous comment, I cannot get headers to work for `Action` even with your changes

denis-reblaze:
> > > @denis-reblaze
> > > I've downloaded this branch and tested it locally, and the headers are not being added to the responses as they should be
> > > According to the rust code found [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/interface.rs#L233) and [here](https://github.com/curiefense/curiefense/blob/main/curiefense/curieproxy/rust/curiefense/src/config/raw.rs#L185) the correct format for headers is `Option<String>`. I'm not sure where you've gotten the `HashMap<String, String>` from, if you could direct me to the matching line that would be appreciated.
> > > In addition, my earlier comment
> > > > Let's use the existing columns instead of creating new ones if possible
> > > 
> > > 
> > > was referring to the bulma `columns` class, not code indentations. We already have a `columns` class separating the entire ResponseAction component into columns and lines, there's no need for another wrapper
> > > edit: formatting
> > 
> > 
> > https://github.com/curiefense/curiefense/blob/ef7ee2916ccf203906472ee639e255f2af70bd30/curiefense/curieproxy/rust/curiefense/src/interface.rs#L120-L129
> 
> That's weird. I'm not sure why there is a discrepancy between the two `Action` types... It looks like the lines I linked were changed from `HashMap<string, string>` to `string`, perhaps there is additional change that is needed to be done in the rust code...
> 
> Anyway, did you test this feature? Does it work for you? Because as I've written in the previous comment, I cannot get headers to work for `Action` even with your changes

No

Aviv-Galmidi:
> > Anyway, did you test this feature? Does it work for you? Because as I've written in the previous comment, I cannot get headers to work for `Action` even with your changes
> 
> No

That's unacceptable! You have to check the functionality of new code you're submitted to the system
I'm converting the PR to a draft for now, please set it as ready for review once the code is working

denis-reblaze:
> > > Anyway, did you test this feature? Does it work for you? Because as I've written in the previous comment, I cannot get headers to work for `Action` even with your changes
> > 
> > 
> > No
> 
> That's unacceptable! You have to check the functionality of new code you're submitted to the system
> I'm converting the PR to a draft for now, please set it as ready for review once the code is working

I don't know how to check

Aviv-Galmidi:
> > > > Anyway, did you test this feature? Does it work for you? Because as I've written in the previous comment, I cannot get headers to work for `Action` even with your changes
> > > 
> > > 
> > > No
> > 
> > 
> > That's unacceptable! You have to check the functionality of new code you're submitted to the system
> > I'm converting the PR to a draft for now, please set it as ready for review once the code is working
> 
> I don't know how to check

Next time if you don't know how to check if a new feature works or a bug is fixed, please either contact me for assitance or refer to the original bug/feature issue

There are a couple of ways to test this
* You can build an image from your local branch and run docker-compose with your images
* You can use your local dev depolyment and save the entity in the new format, then use postman or curl to save it to a running docker-compose and publish the changes (the GUI would not be able to present the data in the new format but that's okay as you can see that it is displaying it correctly on your development deployment)

curiefense/ui/src/components/ResponseAction.vue
denis-reblaze: @Aviv-Galmidi what about this? Implement as shown on the screenshots?

denis-reblaze: @Aviv-Galmidi sorry, forgot that we decided not to implement validation

Issue #485, opened 2021-07-18
upstream.remoteaddress field contains port


flaper87:
**Describe the bug**
Sometimes, it seems like the `upstream.remoteaddress` field contains the port number too:

```
2021-07-18T05:29:42.465Z        WARN    [elasticsearch] elasticsearch/client.go:408     Cannot index event publisher.Event{Content:beat.Event{Timestamp:time.Time{wall:0x2356e8db, ext:63762182974, loc:(*time.Location)(nil)}, Meta:null, Fields:{"agent":{"ephemeral_id":"4ce6f9dd-9f0b-4cb1-ae18-8ae1325acf63","hostname":"c
urie-curielogger-filebeat-filebeat-m2m6d","id":"5746a9ad-918d-4f8a-b1b6-2edda825535a","name":"curie-curielogger-filebeat-filebeat-m2m6d","type":"filebeat","version":"7.10.1"},"blocked":false,"container":{"id":"4dc68ce8157102a5ee915c2765e6f110607e501ee0c85b40ced847deb4aaf20c","image":{"name":"curiefense/curielogger:mai
n"},"runtime":"containerd"},"downstream":{"connectiontermination":false,"directremoteaddress":"","directremoteaddressport":53468,"localaddress":"10.60.4.108","localaddressport":443,"protocolerror":false,"remoteaddress":"151.66.53.16","remoteaddressport":53468},"ecs":{"version":"1.6.0"},"fields":{"cluster":"security-au
tomation"},"host":{"name":"curie-curielogger-filebeat-filebeat-m2m6d"},"input":{"type":"container"},"kubernetes":{"container":{"image":"curiefense/curielogger:main","name":"curielogger"},"labels":{"app_kubernetes_io/name":"curielogger","pod-template-hash":"557596b947"},"namespace":"curiefense","node":{"name":"gke-secu
rity-automat-security-automat-0e0454d2-8wp4"},"pod":{"name":"curielogger-557596b947-9p6cj","uid":"98d15636-9356-47da-aab0-d3d2320fc128"},"replicaset":{"name":"curielogger-557596b947"}},"log":{"file":{"path":"/var/log/containers/curielogger-557596b947-9p6cj_curiefense_curielogger-4dc68ce8157102a5ee915c2765e6f110607e501
[...]

PR #496, opened 2021-07-20
Clean remoteaddress when logging nginx requests

bartavelle:
Also make the nginx container more resistant to restarts.

Should help with #485 


Aviv-Galmidi (approved):

flaper87 (approved):
Thanks

PR #498, opened 2021-07-20
Log the git version of the curiefense binary

bartavelle:


bartavelle:
This PR does not work with the docker builders !

PR #499, opened 2021-07-20
Fix the docker images build on Mac

bartavelle:

### DIFF
--- a/curiefense/images/build-docker-images.sh
+++ b/curiefense/images/build-docker-images.sh
@@ -42,8 +42,9 @@ then
             image=curiefense-rustbuild-${distro}
             IMG=${REPO}/${image}
             echo "=================== $IMG:$DOCKER_TAG ====================="
-            if tar -C curiefense-rustbuild -ch --exclude='.*/target' --exclude='.*/ctarget' . \
-                    | docker build --build-arg UBUNTU_VERSION=${distro} -t "$IMG:$DOCKER_TAG" -; then
+            TMPFILE=$(mktemp)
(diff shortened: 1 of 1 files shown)


curiefense/images/build-docker-images.sh
```diff
@@ -60,6 +61,7 @@ then
                 STP="SKIP"
                 GLOBALSTATUS=1
             fi
+            rm "$TMPFILE"
```
flaper87, line 64: I think it may be better for us to use: `trap "rm -f $TMPFILE" EXIT`

This should go right after the `TMPFILE` declaration

bartavelle: This would mean that all temporary files would be kept until the script exits, right?

The problem with this script and error handling is that ... there is basically none. The whole script should fail every time anything fails, as it takes a long time to run and nobody is closely watching its output. In that case `trap` would certainly be the best. However, as there is currently no way (except for a signal) to avoid the `rm`, this minimizes the amount of temporary files.

This is however probably a futile optimization, as the docker images will take 10x more space anyway ...

Do you think it is worth using `trap` anyway?

flaper87: > This would mean that all temporary files would be kept until the script exits, right?

Correct

I think `trap` is almost always preferable. It may be annoying when trying to debug failures but, at that point, the developer can simply comment the trap line.

bartavelle: Should be ok now

bartavelle: So this is actually more complicated than I expected :)

bartavelle: Here we are, trap resolves variables when it is called, not when it is defined.

Issue #495, opened 2021-07-20
Tag Rules adds an empty tag to each request 


yanushg:
**Describe the bug**
ACL adds an empty tag to each request 
**To Reproduce**
1. Go to ACL tab 
2. Create an empty tag and save 
3. publish 
4. send a request 

ACL conf looks ok: {"id": "__default__", "name": "default-acl", "allow": [], "allow_bot": [], "deny_bot": [], "bypass": [], "deny": [], "force_deny": []}

**Expected behavior**
No tag should be added we add an empty tag 

"tags": ["aclname:default-acl", "urlmap:default-entry", "vpn", "tor", "test", "wafname:default-waf", "aclid:--default--", "container:curieproxy", "wafid:--default--", "yana", "ip:31-154-168-166", "urlmap-entry:default", "tag", "geo:israel", "bot", "", "challenge-phase01", "all", "asn:12400", "curieaccesslog"],

bartavelle:
Is that the acl, or is there a tagging rule with an empty tag?

bartavelle:
If you have `jq` installed, you can do something like that to check it quickly:

```bash
curl http://localhost:30080/conf/api/v1/configs/master/d/tagrules/ | jq -S '.[].tags'
```

yanushg:
I do have an empty tag in tag rules - but this is weird because we have a UI validation 
[
  "vpn",
  "bot",
  "tag",
  "test",
  "yana",
  "tor",
  ""
]
BTW it appears twice in tag rules tags - i have only one list with those tags

Aviv-Galmidi:
Editing the title to better reflect the issue - Tag Rule is adding the extra empty tag and not the ACL

Aviv-Galmidi:
Changing milestone 1.4.0 -> 1.4.1 which is set as the next release as 1.4.0 has been released last night

semharel:
Steps to reproduce the behavior:
create a new Tag rule 
in the tags field to write: "api"
      ->  without the  " ,  but add a space at the end of the tag 

action: 503 
in create new section : Header , key= test, val= tag
<img width="263" alt="Screen Shot 2021-07-29 at 12 23 26" src="https://user-images.githubusercontent.com/86234718/127469477-fa3ef1c4-e917-446c-b86a-54dd14e832b8.png">
save and publish 
send a request:  curl -H "test: tag" "http://{$HOST:PORT}/"
go to kibana :in the tags filed in  the request log  appear  2 tags "api" and " "

[...]

PR #510, opened 2021-07-21
Fixed bug where empty tags were being added to entities

Aviv-Galmidi:
Closes https://github.com/curiefense/curiefense/issues/495
Changed autocomplete error message to not include the word 'tag' as they should be generic and not implementation-specific

### DIFF
--- a/curiefense/ui/src/assets/Utils.ts
+++ b/curiefense/ui/src/assets/Utils.ts
@@ -5,6 +5,9 @@ const invalidityClasses = ` has-text-danger has-background-danger-light`
 
 // Validates an input based on given validator (Function / Boolean) and adds necessary classes if input is invalid
 const validateInput = (event: Event, validator: Function | boolean) => {
+  if (!(event instanceof Event)) {
+    return false
+  }
   let className = (event.target as HTMLElement)?.className
   let isValid
   className = className.replace(`${invalidityClasses}`, '')
@@ -97,6 +100,7 @@ const toast = (message: string | HTMLElement, type: ToastType, undoFunction?: ()
 const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () => any) => {
   const element = document.createElement('div')
   let textElement
+  message = message ? message : ''
   if (typeof message === 'string') {
     textElement = document.createElement('span')
     textElement.innerText = message
@@ -118,7 +122,7 @@ const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () =
 }
 
 const removeExtraWhitespaces = (value: string) => {
-  return value.replace(/\s\s+/g, ' ')
+  return value?.replace(/\s\s+/g, ' ') || ''
 }
 
 export default {
--- a/curiefense/ui/src/assets/Utils.ts
+++ b/curiefense/ui/src/assets/Utils.ts
@@ -1,5 +1,5 @@
-import {ToastType} from 'bulma-toast'
 import * as bulmaToast from 'bulma-toast'
+import {ToastType} from 'bulma-toast'
 
 const invalidityClasses = ` has-text-danger has-background-danger-light`
 
@@ -117,6 +117,10 @@ const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () =
   return element
 }
 
+const removeExtraWhitespaces = (value: string) => {
+  return value.replace(/\s\s+/g, ' ')
+}
+
 export default {
   name: 'Utils',
   validateInput,
@@ -124,4 +128,5 @@ export default {
   generateUniqueEntityName,
   downloadFile,
   toast,
+  removeExtraWhitespaces,
 }
--- a/curiefense/ui/src/components/AutocompleteInput.vue
+++ b/curiefense/ui/src/components/AutocompleteInput.vue
@@ -194,8 +194,8 @@ export default (Vue as VueConstructor<Vue & {
           return
         }
         Utils.toast(
-            `Selected tag "${this.currentValue}" is invalid!\n` +
-            `Tags must be at least ${this.minimumValueLength} characters long.`,
+            `Selected value "${this.currentValue}" is invalid!\n` +
+            `Values must be at least ${this.minimumValueLength} characters long.`,
             'is-danger',
         )
         this.currentValue = ''
--- a/curiefense/ui/src/components/TagAutocompleteInput.vue
+++ b/curiefense/ui/src/components/TagAutocompleteInput.vue
@@ -134,7 +134,7 @@ export default Vue.extend({
           this.tagsSuggestionsLoading = false
         },
       })
-      this.buildTagsSuggestionsFromData(response.data)
+      this.buildTagsSuggestionsFromData(response?.data || {})
       this.tagsSuggestionsLoading = false
       if (this.tagsAddedWhileSuggestionsLoading.length > 0) {
         this.addUnknownTagsToDB(this.tagsAddedWhileSuggestionsLoading)
@@ -171,12 +171,12 @@ export default Vue.extend({
     },
 
     tagChanged(newTag: string) {
-      this.tag = newTag
+      this.tag = newTag.replace(/\s\s+/g, ' ')
       this.$emit('tag-changed', this.tag)
     },
 
     tagSubmitted(newTag: string) {
-      this.tag = newTag
+      this.tag = newTag.replace(/\s\s+/g, ' ')
       // if submitting a tag we don't recognize -> add it to the DB
       if (!this.tagsSuggestions.find((suggestion) => {
         return suggestion.value.toLowerCase() === this.currentTag.toLowerCase()
--- a/curiefense/ui/src/components/TagAutocompleteInput.vue
+++ b/curiefense/ui/src/components/TagAutocompleteInput.vue
@@ -172,12 +172,12 @@ export default Vue.extend({
     },
 
     tagChanged(newTag: string) {
-      this.tag = Utils.removeExtraWhitespaces(newTag)
+      this.tag = Utils.removeExtraWhitespaces(newTag).trim()
       this.$emit('tag-changed', this.tag)
     },
 
     tagSubmitted(newTag: string) {
-      this.tag = Utils.removeExtraWhitespaces(newTag)
+      this.tag = Utils.removeExtraWhitespaces(newTag).trim()
       // if submitting a tag we don't recognize -> add it to the DB
       if (!this.tagsSuggestions.find((suggestion) => {
         return suggestion.value.toLowerCase() === this.currentTag.toLowerCase()
--- a/curiefense/ui/src/components/TagAutocompleteInput.vue
+++ b/curiefense/ui/src/components/TagAutocompleteInput.vue
@@ -25,6 +25,7 @@ import AutocompleteInput, {AutocompleteInputEvents, AutocompleteSuggestion} from
 import Vue from 'vue'
 import {AxiosResponse} from 'axios'
 import {TagsDatabaseDocument} from '@/types'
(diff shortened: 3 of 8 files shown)


curiefense/ui/src/components/TagAutocompleteInput.vue
```diff
@@ -134,7 +134,7 @@ export default Vue.extend({
           this.tagsSuggestionsLoading = false
         },
       })
-      this.buildTagsSuggestionsFromData(response.data)
+      this.buildTagsSuggestionsFromData(response?.data || {})
```
denis-reblaze, line 137: I suggest to refactor the RequestsUtils.sendRequest function so as to always have a response defined (by adding `return {}` at the end of the catch block, or even `return {data: {}}`, or even `return {status: 500, data: {}}`)

curiefense/ui/src/components/AutocompleteInput.vue
```diff
@@ -194,8 +194,8 @@ export default (Vue as VueConstructor<Vue & {
           return
         }
         Utils.toast(
-            `Selected tag "${this.currentValue}" is invalid!\n` +
-            `Tags must be at least ${this.minimumValueLength} characters long.`,
+            `Selected value "${this.currentValue}" is invalid!\n` +
+            `Values must be at least ${this.minimumValueLength} characters long.`,
```
denis-reblaze, line 198: But you told toasts is not the mechanism to display errors, why do we have it here?

curiefense/ui/src/components/TagAutocompleteInput.vue
```diff
@@ -171,12 +171,12 @@ export default Vue.extend({
...
 
     tagChanged(newTag: string) {
-      this.tag = newTag
+      this.tag = newTag.replace(/\s\s+/g, ' ')
       this.$emit('tag-changed', this.tag)
     },
 
     tagSubmitted(newTag: string) {
-      this.tag = newTag
+      this.tag = newTag.replace(/\s\s+/g, ' ')
```
denis-reblaze, line 179: Duplicates line 174, should be implemented as a separate function

curiefense/ui/src/components/TagAutocompleteInput.vue
Aviv-Galmidi: I do not agree that sendRequest should always return an object with data, when the server returns status 500. Perhaps we should return response.data directly from RequestUtils but that's
Anyhow, that's another discussion all together, this line is a hotfix as the tests were not being run for this component after the latest changes for RequestUtils

curiefense/ui/src/components/AutocompleteInput.vue
Aviv-Galmidi: A couple of things
1) I said toasts are not a mechanism to display *temporary* errors, this is an attempt to submit a tag(or value) which results in a permanent error - we do not apply the new tag and remove it from the input. Unlike field validation, you can't change the value again as it is being deleted. This error is closer to the "failed to save" error than field validation errors
2) I still agree that it is not the optimal way of displaying the error, the error should be directly next to the error-prone field. However, this all area of code is going to be modified soon as we move the tag creation process to the new system. This change was because AutoCompleteInput shouldn't know what "tags" are as they're a single implantation of the auto complete and thus this message should be more generic to all usecases

curiefense/ui/src/components/TagAutocompleteInput.vue
Aviv-Galmidi: Fixed, moved to a new function in Utils.ts

curiefense/ui/src/assets/Utils.ts
```diff
@@ -117,11 +117,16 @@ const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () =
   return element
 }
 
+const removeExtraWhitespaces = (value: string) => {
+  return value.replace(/\s\s+/g, ' ')
```
denis-reblaze, line 121: The function will throw an exception if the value is, for example, undefined

denis-reblaze (changes requested):

curiefense/ui/src/components/TagAutocompleteInput.vue
Aviv-Galmidi: I did not say it is better than a single `return {}` but you did not suggest a signle `return {}` you suggested to `return {data: {}}`
If you re-read my comment you'll see I had a problem with the `data: {}` part as responses with server status 500 shouldn't have data attribute in them

curiefense/ui/src/components/AutocompleteInput.vue
Aviv-Galmidi: Of course they are technical terms, and they have technical meanings as well
We display some error in toast messages as they are always true - the server request failed, sending it again does not change the fact that the inital request failed. Attempting to submit a tag with 2 characters fails, further attempts do submit a different tags with more characters do not change this fact.
We do not display temporary errors in toasts - the input is currently not a valid email, you can change it and it will be a valid email, it is still the same input data, so once you change it you'd need to cancel out the error which cannot be done with toasts (technically can be done, but that's not how toasts perform, they're a messaging system)

curiefense/ui/src/assets/Utils.ts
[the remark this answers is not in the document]
Aviv-Galmidi: There is little to no benefit to writing it in a single line
Multiple lines are more readable and are easier to maintain, you're saving a few bytes of data but that's negligible

curiefense/ui/src/components/TagAutocompleteInput.vue
denis-reblaze: Seems that rather you need to re-read my comment: `return {}` was the very first suggestion.

curiefense/ui/src/assets/Utils.ts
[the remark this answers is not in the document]
denis-reblaze: The benefit is not in bytes but in saving a developer's eyes, and therefore increasing the development quality. Unnecessary `return`s here and there are clogging the code (btw, this was one of the goals of inventing arrow functions, why not to use it?)

Aviv-Galmidi: Fixed for all utils functions

curiefense/ui/src/components/TagAutocompleteInput.vue
Aviv-Galmidi: I stand corrected, you sugged returning an empty object, but that's really not the point. All I've addressed is the fact that I do not think sendRequest should return data on errors, and we should hold a discussion to decide on the best implementation later on. I did not suggest adding `response?.data || {}` at every place where we implement sendRequest
I merely added a hotfix in order to fix the tests suite that was broken with the latest changes to sendRequests. I do not know the implications of changing the catch to return an object at the moment, and i do not have the capacity to run full manual tests for such a change before the release of 1.4.0

curiefense/ui/src/assets/Utils.ts
```diff
@@ -97,6 +100,7 @@ const toast = (message: string | HTMLElement, type: ToastType, undoFunction?: ()
 const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () => any) => {
   const element = document.createElement('div')
   let textElement
+  message = message ? message : ''
```
denis-reblaze, line 103: It's better to use default function parameter or another variable

curiefense/ui/src/components/AutocompleteInput.vue
denis-reblaze: I got this idea from the previous comment, and the suggestion was to get rid of the technical meanings in order to look at the system as an ordinary user.

curiefense/ui/src/assets/Utils.ts
```diff
@@ -117,11 +121,16 @@ const buildToastUndoElement = (message: string | HTMLElement, undoFunction: () =
   return element
 }
 
+const removeExtraWhitespaces = (value: string) => {
+  return value?.replace(/\s\s+/g, ' ') || ''
```
denis-reblaze, line 125: still potentially crushable if somehow the value's type won't be string

curiefense/ui/src/components/TagAutocompleteInput.vue
denis-reblaze: ok, I hope this function will be refactored asap, when we have time

denis-reblaze (changes requested):

curiefense/ui/src/assets/Utils.ts
Aviv-Galmidi: Default function parameters do not overwrite null values, so that wouldn't work
Why do you believe a new vairable would help here?

curiefense/ui/src/assets/Utils.ts
Aviv-Galmidi: That's what typing is for in typescript, we don't need to check if value is a string as we require it to be typed as string

curiefense/ui/src/assets/Utils.ts
denis-reblaze: Correct, default parameter is not an option. About immutability it's written a lot, the main benefit is 

> simpler programming and debugging, as data that never changes is easier to reason about than data that is free to be changed arbitrarily throughout the app.

curiefense/ui/src/assets/Utils.ts
denis-reblaze: Ts helps on the dev step only. You're using the function in a component's event, which was emitted from another component - can you be sure that types are matched in this and the other cases? If yes, I'm calling this comment off, but in that case you need to remove the recently added checking `value? ||`.

Aviv-Galmidi: `value` is a string, it cannot be anything else other than `string`, `null`, or`undefined`. The `value?... || ''` check is in case the data sent to the function is either `undefined` or `null` in which case `.replace` would return an error

curiefense/ui/src/components/AutocompleteInput.vue
Aviv-Galmidi: We are using toasts when they are applicable and we do not use toasts when they are not.

A toast can show an error if a user completed an action, but there was a system-related issue preventing the action from being fully executed
Of course we'll bind the technical meaning of an error to the usage of toasts and displaying of errors, because that's they are differrent types of errors. A system-related error in response to a user-submitted action is not the same type of error as a form field not complying with a pre defined set of rules

What you're suggesting is similar to this: `Let's remove button colors as positive buttons (blue) and negativie buttons (red) have the same meaning once we remove their technical meaning`. They have different technical meanings and thus different display patterns

curiefense/ui/src/assets/Utils.ts
Aviv-Galmidi: You're taking a qoute out of context, as this qoute is refering to react + redux change detection. If you want to bring a qoute you should bring the entire qoute.

> Immutability can bring increased performance to your app, and leads to simpler programming and debugging, as data that never changes is easier to reason about than data that is free to be changed arbitrarily throughout your app.
In particular, **immutability in the context of a Web app enables sophisticated change detection techniques to be implemented simply and cheaply** (emphasis mine), ensuring the computationally expensive process of updating the DOM occurs only when it absolutely has to (a cornerstone of React’s performance improvements over other libraries).

Javascript is using pass-by-value (in this case) which means this string will always be inclosed within this function and will only live inisde of it. The change will never have any ramifications on the change-detection process.

We do not have the risk of not understanding the change down the line, it is a simple default state in case `message` is `undefined` or `null` and it is like writing the following
```
if (!message) {
    message = ''
}
```

flaper87 (approved):

PR #539, opened 2021-08-05
Fix curieconf unit tests

Freakachoo:
# Change log
- fix unit tests and mock data of `curieconf`
- fix limit model to support arrays for `include` and `exclude` properties. The initial change was made in:  https://github.com/Freakachoo/curiefense/commit/40422d49a1be269d579c296bee7ed844de22b0bd  and https://github.com/Freakachoo/curiefense/commit/0b719967c170a0b364f7082a03df6ba3a472db47

@flaper87 could you fix please the Ci/CD `Run curieconf Python tests` action? At the moment it is running tests as:
```
pytest "--junitxml=$GITHUB_WORKSPACE/junit.xml" --suppress-tests-failed-exit-code
```
I think this prevents the falling of the action even if there were failed tests. Now tests are fixed and we may enable them again, so we may watch on code changes and keep tests updated. Thank you!


curiefense/curieconf/server/curieconf/confserver/v1/api.py
```diff
@@ -48,8 +48,8 @@ class AnyType(fields.Raw):
         "ttl": fields.String(required=True),
         "limit": fields.String(required=True),
         "action": fields.Raw(required=True),
-        "include": fields.Raw(required=True),
-        "exclude": fields.Raw(required=True),
+        "include": fields.List(fields.String()),
+        "exclude": fields.List(fields.String()),
```
shaisgv, line 52: Please check if this is a required value. If its not mandatory, consider moving it to the end.

shaisgv (changes requested):

shaisgv (approved):

PR #553, opened 2021-08-10
Add an option to upload configuration through the GUI

denis-reblaze:
We would like to give the user the ability to upload configuration through the GUI
The new feature will act in the same manner as the Download buttons around the system, but instead of downloading a JSON configuration file, it would upload the configuration

In every location where we have a Download button, add a new button called "Upload" next to it with the functionality described above. The uploaded data would override the existing data whenever uploaded

screenshot-localhost_8080-2021 08 10-17_09_12

### DIFF
--- a/curiefense/ui/src/assets/Utils.ts
+++ b/curiefense/ui/src/assets/Utils.ts
@@ -1,5 +1,6 @@
 import * as bulmaToast from 'bulma-toast'
 import {ToastType} from 'bulma-toast'
+import {AxiosResponse} from 'axios'
 import {GenericObject} from '@/types'
 
 const invalidityClasses = ` has-text-danger has-background-danger-light`
@@ -59,10 +60,7 @@ const downloadFile = (fileName: string, fileType: string, data: any) => {
     return
   }
   // Create downloadable content based on file type
-  let content: BlobPart = ''
-  if (fileType === 'json') {
-    content = JSON.stringify(data)
-  }
+  const content: BlobPart = JSON.stringify(data)
   // Create anchor element with download data
   const blob = new Blob([content], {
     type: `application/${fileType}`,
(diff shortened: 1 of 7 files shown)


curiefense/ui/src/assets/Utils.ts
```diff
@@ -73,6 +72,30 @@ const downloadFile = (fileName: string, fileType: string, data: any) => {
   link.click()
 }
 
+type UploadFileFunction = Promise<{response: Partial<AxiosResponse>, uploadData: GenericObject[]}>
+const uploadFile = (file: File, dataSender: Function, dataValidator: Function): UploadFileFunction => {
+  return new Promise((resolve) => {
+    const reader = new FileReader()
+    const fileType = file.name.split('.')[1]?.toLowerCase()
```
Aviv-Galmidi, line 79: file name might have multiple dots in the name. e.g. `this.is.a.file.json` or an example from our own codebase `vue.config.js`
Let's take the last item in the split array instead of the second one

Please also add a test to cover this scenario

curiefense/ui/src/assets/Utils.ts
```diff
@@ -73,6 +72,30 @@ const downloadFile = (fileName: string, fileType: string, data: any) => {
   link.click()
 }
 
+type UploadFileFunction = Promise<{response: Partial<AxiosResponse>, uploadData: GenericObject[]}>
```
Aviv-Galmidi, line 75: This isn't the type of the function but the response type, the name is somewhat misleading

```suggestion
type UploadFileFunctionResponse = Promise<{response: Partial<AxiosResponse>, uploadData: GenericObject[]}>
```

Aviv-Galmidi (changes requested):

PR #560, opened 2021-08-15
e2e tests fix

Freakachoo:
# Changelog
- setup utils separately, before tests run and dockers built


Freakachoo:
Tests showed that issue were just in "include" "exclude" json

PR #563, opened 2021-08-18
Always set directremoteaddress in ngx lua

bartavelle:
If not, it is set as an empty string in curielogger, which prevents
logging from working.


flaper87 (approved):
:pizza:

PR #567, opened 2021-08-23
Prevent ELK from fail on GitHub Actions

Freakachoo:
Sometimes on GitHub actions ELK is falling and prevents tests to finish.
Here is the possible solution: https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch
Let's add it and hope it helps.

# Change log 
Add sysctl settings for GitHub actions VMs to help ELK not to fail

### DIFF
--- a/.github/workflows/e2e-compose-tests.yml
+++ b/.github/workflows/e2e-compose-tests.yml
@@ -20,6 +20,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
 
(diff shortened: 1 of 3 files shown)


flaper87 (changes requested):
What do you think about putting it in the bash scripts instead? That way we have the settings in one place only

Freakachoo:
> What do you think about putting it in the bash scripts instead? That way we have the settings in one place only

That's ok, but the funny thing - it did not prevent tests from failing :))) I suggest just reject this change if it doesn't work anyway.

Freakachoo:
The change didn't help to prevent ELK from falling. Canceling.

PR #568, opened 2021-08-23
Change term ACLPolicy to ACLProfile

Freakachoo:
# Change log
- changed `aclpolicy` to `aclprofile` everywhere


flaper87:
Won't this change break backwards compatibility for the API?

Freakachoo:
> Won't this change break backwards compatibility for the API?

@flaper87 it will but only in terms of migration from `v1` configs version to `v2` configs. But we're gonna create:
1. Script to migrate from `v1` to `v2`
2. Instructions to do this (include to docs)

Wanted your advice, btw - should we do it as a script+manual instructions or automatic conversion? What's the procedure should be to do rolling updates from `v1` to `v2`? Maybe with using different `buckets`, or different branches for configs? Would want to discuss.

Issue #552, opened 2021-08-10
Terminology & Hierarchy changes


Freakachoo:
# Issue motivation 

### Policies & Rules hierarchy
Different document types:
- Tag Rules -> (Profiling Filter?)
- Flow Control
- Rate Limit Rules
- ACL Policies
- WAF Policies
- WAF Rules
- URL Maps

### Current state:
- URL Map (containing many ACL Policies, WAF Policies, Rate Limit Rules)
- ACL Policies (containing many Tags)
- WAF Policies (containing many WAF Rules)
  - WAF Rules (flat)
- Rate Limit Rules
- Flow Control (containing many tags)
- Tag Rules (flat)

There is no consistency between the names and their respective hierarchy levels
We have two options that were talked about regarding Rate Limit Rules
1. Create a new entity called “Rate Limit Policy” which would contain multiple “Rate Limit Rules”, this policy would be used in URL Maps.
This approach matches the approach we have with ACL Policy (which contains multiple tags or ACLs) and WAF Policy (which may contain multiple WAF Rules)
2. Change Rate LImit Rule to assign a tag to each request which matches its parameters, and use the newly created tags in URL Maps instead of Rate Limit IDs
This approach keeps us in the “world of tags” and was suggested by the users, however it’s a new kind of tag which does not exactly match the current tags and we would need to maintain multiple tag lists

### Policies & Rules terms

1. Fields “notes” and “description” act the same
“Notes” used in Tag Rules & Flow Control
“Description” used in Rate Limit Rules
These fields are comments for use within the interface
	
2. Field “TTL” is not informative
“TTL” used in Flow Control & Rate Limit Rules
This field is described differently for each of those rules
Flow Control - The time period within which the traffic source must complete the Flow Control Sequence
Rate Limit Rules - the period within which the Threshold is enforced
Proposed new name “Timeframe”

3. “Count by” field is not informative
“Count by” used in Flow Control & Rate Limit Rules
This field describes the matching conditions and criteria by which Curiefense will associate requests with a single requestor
“Count by” has two fields “Type” and “Key”
	- “Type” - Header/Cookie/Argument/Attribute
	- “Key” - Key/Value for matching type to count by

4. ACL Policies different tag lists names have little meaning
Enforce Deny, Bypass, Allow Bot, Deny Bot, Allow, Deny - These names do not convey their meaning in an easy to understand way
In addition to this, we can add descriptive information labels to each list as described in the docs
Proposed names:
- Enforce Deny -> 
[...]

PR #571, opened 2021-08-25
Term change FlowControl to FlowControlPolicy

Freakachoo:
According to issue https://github.com/curiefense/curiefense/issues/552 the term `FlowControl` was changed to `FlowControlPolicy`.

# Changelog
The term was changed for:
- UI

# TODO
Update docs


Freakachoo:
@flaper87 I'm thinking about to make this change only in UI. Don't touch short `flow` in curieproxy and `flowcontrol` in curieconf and paths. What do you think?

flaper87:
> @flaper87 I'm thinking about to make this change only in UI. Don't touch short `flow` in curieproxy and `flowcontrol` in curieconf and paths. What do you think?


Sounds like a better strategy to me! :+1:

Aviv-Galmidi (approved):

Issue #460, opened 2021-07-13
Tag rules - ASN - no auto complete 


yanushg:
**Describe the bug**
ASN attribute has no auto complete 
**To Reproduce**
1. Create a tag rule list with ASN attribute 
2. start typing for auto complete 
**Expected behavior**
we should autocomplete and change the ui for ASN ( see screenshot)  
![Screen Shot 2021-07-13 at 13 21 01](https://user-images.githubusercontent.com/48092909/125435961-695c82c6-3700-46bd-bb86-0c99fc26c966.png)

yanushg:
Country attribute as well

PR #585, opened 2021-09-09
ASN and Country fields as autocompletes

denis-reblaze:
As desribed here: https://github.com/curiefense/curiefense/issues/460
We'd like to add a list of ASNs and Countries and change the implementations of these inputs throughout the system to an autocomplete.

Asns and countries lists placed into corresponding json files in assets, autocomplete component improved to be able to filter huge arrays.

Screen Shot 2021-09-09 at 20 07 07


curiefense/ui/src/components/AutocompleteInput.vue
```diff
@@ -248,6 +251,19 @@ export default (Vue as VueConstructor<Vue & {
       clearTimeout(this.inputBlurredTimeout)
     },
 
+    * filter(array: AutocompleteSuggestion[], condition: Function) { // the fast analogue of Array.filter
```
Aviv-Galmidi, line 254: I think we can use `array.slice(0, this.maxSuggestionsNumber)` to get the first X items. I think it would be easier to read as well as simply 1 line compared to multiple lines

I don't know the performance difference between the two, but as we display ~15 items, I don't think either of the solutions' performance would hinder the system at all

denis-reblaze: I don't need the first X items but the first X items that satisfy `condition`. We cannot use array.slice for that.

Issue #566, opened 2021-08-22
Tag rule ui takes tags from first list 


yanushg:
**Describe the bug**
when creating a tag rule via api the ui takes tag from the first list 
**To Reproduce**
1. go to API 
2. Create a new tag rule with tag "test" 
3. open tag rule you have created in the UI 

**Expected behavior**
Should take the correct tag test but instead takes "api" tag

Aviv-Galmidi:
Can be re-tested, should be fixed now

PR #588, opened 2021-09-19
Tag autocomplete initial value fix

denis-reblaze:
Description: https://github.com/curiefense/curiefense/issues/566

Actually the problem is due to an initial value of the Tagautocomplete component which needs to be watched.


Aviv-Galmidi (approved):

flaper87 (changes requested):
Can you please squash all these commits in a single one? There are too many merge commits that should probably not be there.

Aviv-Galmidi:
I'll be squashing the commits when merging as requested

PR #621, opened 2021-11-12
Convert the ngx image to bionic

bartavelle:
Once nginx-ingress too is converted, we will be able to build only one
Rust.


bartavelle:
Once it is converted, it will be possible to remove the focal rust build from the script.

flaper87 (approved):

PR #627, opened 2021-11-20
deploy-gke.sh: specify cluster version & use JWT WORKAROUND

xavier-rbz:
GKE now runs version 1.21 by default, which causes an authentication error between the ingressgateway and istiod.
Updating istio would probably fix this (see [this PR](https://github.com/istio/istio/pull/33723)), let's do that soon.


flaper87 (approved):
when do you think we can update istio?

PR #628, opened 2021-11-22
uiserver: add config knob to use https for proxied requests to confserver

xavier-rbz:
Potential users have expressed the need for requests between uiserver and confserver to be conveyed over HTTPS rather than HTTP.
This PR adds a new environment variable, `HTTPS_FOR_CONFSERVER`, to the uiserver image, to enable that.
Users can then add an nginx sidecar (or similar) to the confserver deployment to expose it over https, with a certificate & key they provision.

I plan to update the documentation to explain how to do that.


curiefense/images/confserver/init/start_nginx.sh
```diff
@@ -2,4 +2,9 @@
 
 /bootstrap/initial-bucket-export.sh 1>/dev/stdout 2>/dev/stderr &
 
+# Enable TLS if required secrets are present -- k8s environments
+if [ -f /run/secrets/confsslcrt/confsslcrt ]; then
```
flaper87, line 6: I'm a bit hesitant to having paths like this one hard-coded. What do you think about having an env var that has the path to the tls secret? As it is now, the name of the secret and the path where it's mounted are hardcoded, which may not be ideal for folks deploying using other tools

xavier-rbz: Good idea, I'll work on that and update the PR

PR #638, opened 2021-12-07
Asynchronous C FFI for curiefense

bartavelle:
Can be integrated in Web servers that have asynchronous innards.

The change is important, and made of the following parts, that, unfortunately, depend on each other:

 * the smallest change is code reorganization, so that imports make more sense (especially with regards to grasshopper functionalities)
 * all code that interacts with Redis is converted to its asynchronous equivalent, and all functions updated accordingly
 * a very simple stepping executor for asynchronous function is implemented (in `simple_executor.rs`). It might be replaced by something more efficient in the future
 * a C-compatible API is exposed.

The code is tested in the exact same way the previous code was : unit tests, lua driven tests and e2e tests. It has also been tested manually in a (non published) Nginx module.


flaper87 (commented):
This change is massive :/ 

Could you elaborate more in the PR description on what the change is about, pros, cons, how to use it, and how it was tested? :pray: 

If anything, it will serve for history purposes

bartavelle:
> This change is massive :/
> 
> Could you elaborate more in the PR description on what the change is about, pros, cons, how to use it, and how it was tested? pray
> 
> If anything, it will serve for history purposes

I updated the description. Let me know if this is insufficient, or if you'd like more details in the code itself.

PR #647, opened 2021-12-26
Validate Ban Actions

Freakachoo:
# Changelog
- Validate Ban that rate limit can have only one Ban action
- Search Ban action before applying in rust code 
- Fix TESTING to TESTIMG in build script
- Client-side validation - only one Ban action can be selected
image

### DIFF
--- a/curiefense/ui/src/doc-editors/RateLimitsEditor.vue
+++ b/curiefense/ui/src/doc-editors/RateLimitsEditor.vue
@@ -98,7 +98,7 @@
                             type="text"
                             title="A number of requests"
                             placeholder="A number of requests"
-                            @change="updateThresholdOption($event, index)"
+                            @change="emitDocUpdate"
                             v-model="threshold.limit">
                     </div>
                     <div class="button-wrapper-column column">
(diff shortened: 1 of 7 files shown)


curiefense/ui/src/doc-editors/RateLimitsEditor.vue
```diff
@@ -484,6 +497,14 @@ export default Vue.extend({
       this.checkKeysValidity()
     },
 
+    checkOnlyOneBan() {
```
Aviv-Galmidi, line 500: We can make this into a computed property instead of calculating it ourselves each time

Aviv-Galmidi (changes requested):

curiefense/ui/src/doc-editors/RateLimitsEditor.vue
Freakachoo: Done! Re-review please

flaper87 (approved):
Rust bits and limits look good to me :)

curiefense/ui/src/doc-editors/RateLimitsEditor.vue
```diff
@@ -462,6 +474,10 @@ export default Vue.extend({
       this.emitDocUpdate()
     },
 
+    updateThresholdOption(option: OptionObject, index: number) {
```
Aviv-Galmidi, line 477: We can remove this function now and call emitDocUpdate directly

Aviv-Galmidi (approved):

PR #654, opened 2022-01-03
Let multiple limits work at the same time

### LINKED ISSUE #650: Security policy second added rate limit is not enforced
**Describe the bug**
When adding 2 rate limit rules to security policy the second rule that was added is not enforced.
**To Reproduce**
1. configure 2 rate limit rules
2. Rule A with time frame of 30 sec and count by header (any value) with threshold 3 and action 503 service unavailable
3. Rule B with time frame of 30 sec and count by header (a different value) with threshold 4 and action response with any values
4. both rules will check by http request event.
5. in your security policy add both rules. add rule a first and rule b second
6. publish the changes
7. sent requests the meets rule A - verify traffic is blocked according to the threshold
8. sent requests the meets rule B - verify traffic is blocked according to the threshold

rule B traffic is not blocked. When adding rule B first and rule A second to security policy rule A traffic is not blocked. 

Seems that when adding 2 rukes only the first rule that is added is enforced. 

**Desktop (please complete the following information):**
 - OS: Mac os
 - Browser chrome
 - Version 1.4.0

**Additional context**
[saar.csv](https://github.com/curiefense/curief
[logger.log](https://github.com/curiefense/curiefense/files/7788501/logger.log)
ense/files/7788483/saar.csv)
[saar.csv.zip](https://github.com/curiefense/curiefense/files/7788505/saar.csv.zip)

bartavelle:
Fixes #650. The bug was that when a "limit key" can't be built, the
filter decision would be to let pass instead of checking the other
limits.

This means that, for exemple, if a limit counts on a header, and that
header was not present, then subsequent limits would not be checked.

### DIFF
--- a/curiefense/curieproxy/rust/curiefense/src/limit.rs
+++ b/curiefense/curieproxy/rust/curiefense/src/limit.rs
@@ -130,7 +130,9 @@ pub fn limit_check(
         }
 
         let key = match build_key(security_policy_name, reqinfo, limit) {
-            None => return SimpleDecision::Pass,
+            // if we can't build the key, it usually means that a header is missing.
+            // If that is the case, we continue to the next limit.
+            None => continue,
             Some(k) => k,
         };
         logs.debug(format!("limit={:?} key={}", limit, key));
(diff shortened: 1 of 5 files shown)


xavier-rbz (approved):
Looks good. I'll add that to the e2e tests later!

flaper87 (approved):

PR #655, opened 2022-01-03
Order security policies by decreasing matcher length

bartavelle:
So that more specific rules get selected first.


bartavelle:
Should fix #652

Aviv-Galmidi (approved):

bartavelle:
Hum perhaps I should have merged it with dev? Not sure I did the right thing here

flaper87:
> Hum perhaps I should have merged it with dev? Not sure I did the right thing here

We only merge on main, fwiw

PR #656, opened 2022-01-03
Improve acl schema

semharel:
Improve the validation of the ACL profile scheme
Eliminate the use of double schemas (the JSON schema file  and model of flask_restx)
get out the ACL model to a separate file (curie_models)


curiefense/curieconf/server/curieconf/confserver/v2/api.py
```diff
@@ -3,19 +3,15 @@
...
+from flask_restx import Resource, fields, marshal, reqparse
 from curieconf import utils
 from curieconf.utils import cloud
 import requests
 from jsonschema import validate
 from pathlib import Path
 import json
-
-
-api_bp = Blueprint("api_v2", __name__)
```
Aviv-Galmidi, line 16: Why did you delete this part? It's necessary in order to define all of the models

curiefense/curieconf/server/curieconf/confserver/v2/api.py
```diff
@@ -618,9 +597,13 @@ def post(self, config, document):
...
-        data = marshal(request.json, models[document], skip_none=True)
-        res = current_app.backend.entries_create(config, document, data)
-        return res
+        isValid, reason = validateJson(request.json, document)
+        if isValid:
+            data = marshal(request.json, models[document], skip_none=True)
+            res = current_app.backend.entries_create(config, document, data)
+            return res
+        else:
+            abort(500, "schema mismatched because the {}".format(reason))
```
Aviv-Galmidi, line 606:
```suggestion
            abort(500, "schema mismatched because of the following reason(s): {}".format(reason))
```

curiefense/curieconf/server/curieconf/confserver/v2/api.py
```diff
@@ -636,13 +619,14 @@ def put(self, config, document, entry):
...
-        isValid = validateJson(request.json, document)
+        isValid, reason = validateJson(request.json, document)
+
         if isValid:
             data = marshal(request.json, models[document], skip_none=True)
             res = current_app.backend.entries_update(config, document, entry, data)
             return res
         else:
-            abort(500, "schema mismatched")
+            abort(500, "schema mismatched because the {}".format(reason))
```
Aviv-Galmidi, line 629:
```suggestion
            abort(500, "schema mismatched because of the following reason(s): {}".format(reason))
```

Aviv-Galmidi (changes requested):
Please fix the CodeQL issues as well as DCO and any tests that broke during this change
In addition, please also remove the unnecessary whitespace you've added to the files

curiefense/curieconf/server/curieconf/confserver/v2/api.py
semharel: I moved it to the new file ״curie_models״

curiefense/curieconf/server/curieconf/confserver/v2/api.py
semharel: Done

curiefense/curieconf/server/curieconf/confserver/v2/api.py
semharel: Done

curiefense/curieconf/server/curieconf/confserver/v2/api.py
Aviv-Galmidi: The API is not part of the model, the model is part of the API. Therefore, we can't declare the API in the `curie_models` file

We'll need to either return the model here as it is written directly on the `api` variable, move the `api` variable to a new third file, or find another solution all together

curiefense/curieconf/server/curieconf/confserver/v2/curie_models.py
```diff
@@ -0,0 +1,315 @@
+from flask_restx import fields, Namespace
+from curieconf import utils
+
+api = Namespace("models")
```
Aviv-Galmidi, line 4: You're still declaring the `api` variable inside of `curie_models` which is not the right place for it, the `api` is not part of the model decleration, the models are part of the `api`

PR #661, opened 2022-01-05
Performance testing with locust

xavier-rbz:
Locust runs several "workers" that issues queries, which allows us to
reach higher request per second (RPS) than with fortio. Queries sizes are
now a parameter; the test script uses this new feature to collect and
generate report.

When testing with locust, we deploy 3 nodes:
* 1 node for locust master & workers that generate traffic
* 1 node for istio's ingressgateway
* 1 node for all other components: confserver, target services

This PR uses minio to store published configurations instead of S3,
which means AWS credentials are not needed anymore.

### DIFF
--- a/deploy/curiefense-helm/example-miniocfg.yaml
+++ b/deploy/curiefense-helm/example-miniocfg.yaml
@@ -1,0 +1,22 @@
+---
+apiVersion: v1
+kind: Secret
+data:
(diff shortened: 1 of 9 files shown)


flaper87 (commented):
What do you think about adding the minio files in a different PR?

xavier-rbz:
This PR was superseded by #803 (now merged), closing.

PR #664, opened 2022-01-10
Move the `eu` field from `country` to its parent container, `geo`

### LINKED ISSUE #663: Logging failure when source IP is from an EU country
In the curielogger code, in the `Geo` structure, country is defined as:

```go
Country   map[string]string 
```

However, the country object looks like:

```json
    "country": {
      "eu": true,
      "iso": "fr",
      "name": "france"
    }
```

The `eu` field is a boolean, which prevents the message from being logged.

bartavelle:
Shoud fix #663 


flaper87 (approved):
No changes needed to the `lua` bits? Interesting

bartavelle:
New version : the `eu` field is now part of the Geo field. `curielogger` currently filters it out, but at least part of it is logged now!

bartavelle:
> No changes needed to the `lua` bits? Interesting

This part of the message is directly forwarded from the Rust code, so no alteration should be necessary.

flaper87 (changes requested):
I think this PR needs to be re-titled and we now need changes to curielogger for it to work since the EU field needs to be moved there as well.

bartavelle:
Renamed. With this PR, curielogger doesn't reject the message, but indeed filters the new `eu` field out.

flaper87:
> Renamed. With this PR, curielogger doesn't reject the message, but indeed filters the new `eu` field out.

Yeah, that's what I meant. Could you add it to this PR too? Should be enough to add it to [this struct](https://github.com/curiefense/curiefense/blob/main/curiefense/curielogger/pkg/entities/misc.go#L83)

flaper87 (approved):

PR #696, opened 2022-02-02
Geoip regions - Rust part

bartavelle:
This builds on top of #695 , so it should be applied first.

I added code to load the new GeoIP database files, but I am unsure if this is necessary.

What is missing for this feature to work is:
 * UI support
 * logger support


flaper87:
> I added code to load the new GeoIP database files, but I am unsure if this is necessary.

Can you elaborate on what these new GeoIP files are for, the difference in format, and why this PR is needed? :pray: 

(just for the sake of understanding and history)

PR #716, opened 2022-02-17
Make names of folders that store configs more specific

xavier-rbz:
Two types of configs are currently stored in a folder called /config:
1. in confserver pods, to store the persistent git repo that contains
   confserver's data
2. in curieproxy pods, to store generated configurations (json and
   maxmind files), where they are updated by curieproxy and read by the
   envoy rust module

This commit aims at fixing two issues:
* the use of the "config" name, which is too common. Curiefense users
  have shared deployment customizations where they were already using
  /config to store other sorts of configurations, which conflicted with
  curiefense's use
* the use of the same folder name to store very different things, which
  can be confusing

To achieve this, this commits changes folder names:
* persistent data for confserver will now be stored in
  /cf-persistent-config
* generated configs for the envoy rust module will be in /cf-config

### DIFF
--- a/curiefense/curieconf/server/app/main.py
+++ b/curiefense/curieconf/server/app/main.py
@@ -1,4 +1,4 @@
 from curieconf.confserver import app
 from curieconf.confserver.backend import Backends
 
-app.backend = Backends.get_backend(app, "git:///config/confdb")
+app.backend = Backends.get_backend(app, "git:///cf-persistent-config/confdb")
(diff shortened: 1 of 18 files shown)


xavier-rbz:
e2e tests on minikube fail, because Helm charts use `/config` instead of the new `/cf-config`. [This PR](https://github.com/curiefense/curiefense-helm/pull/19) fixes the chart. I have manually deployed the updated chart on GKE, and confirmed that:
* confserver's bootstrap configuration is created properly
* configuration updates are propagated to curieproxy and picked up by the envoy rust filter, which applies policies as expected.

Both PR's (the current PR and [the corresponding Helm chart changes](https://github.com/curiefense/curiefense-helm/pull/19)) should be merged simultaneously. The [published Helm chart](https://helm.curiefense.io/) should be updated as well.

xavier-rbz:
Thanks for the review!
Depending on the container (confserver vs curieproxy/curiesync), `/config` currently contains completely different things.
The reason I want to take this opportunity to use different names, is that using different names for different things would reduce confusion. It would make it clearer for users that confserver's persistent data store (~a git repo) is different from exported configurations that are read by curieproxy.

xavier-rbz:
This was merged as part of #783 ([commit](https://github.com/curiefense/curiefense/pull/783/commits/4fa735740bd92f2fac9924eecb4a05ad0277313d)), closing.

PR #726, opened 2022-02-21
Fix logging when human & bot are ACL blocked

bartavelle:
The blocking logic was sound, but the logging would not display the
proper block reason when bot + human were set to block.

This patch adds a special case.


flaper87 (approved):
:shipit:

PR #732, opened 2022-02-22
Waf risk level rust

bartavelle:
"Rust only" part of the waf risk level patch


flaper87:
> This is not to be merged as if, right now.

you can switch the PR to `Draft` if it's not ready for review. Another option is using `WIP` (work in progress)  in the title so that reviewers know that there's no need to review the PR just yet :)

bartavelle:
This has been merged!

PR #735, opened 2022-02-23
Ignore unexisting pairwith selectors in rate limit

bartavelle:
Previously, failing selectors (for example when matching on a
nonexisting header) for the `pairwith` configuration would act as if
there was no `pairwith` configuration set.

This patch now ignore limits where `pairwith` can't be resolved.


flaper87 (approved):
:shipit:

PR #736, opened 2022-02-23
Add the capacity to select by tags

bartavelle:
Tags are sorted and aggregated in order to create a string.

This can be used as counting for rate limiting, with the effects that
requests with identical tags are lumped together.

However, this is expected to be flaky as tags are altered during the
rate limiting operation ...


flaper87 (approved):
:shipit:

flaper87 (approved):

PR #755, opened 2022-03-10
Fix waf bypass bug

bartavelle:
In waf policies, a regex can be set to match values of arguments. When
this value is empty in the UI, it is stored as an empty string in the
configuration file. In that case, the resulting regex would be the empty
regex, matching everything.

This patch fixes that bug, and adds tests.


bartavelle:
note: included in the `merge` branch

flaper87:
> note: included in the merge branch

not sure I follow this comment :D

flaper87 (approved):

PR #769, opened 2022-03-20
Fix bug where Security Policy save button being disabled at times

Aviv-Galmidi:
Fix bug where initial domain is not correctly set in some occasions in SecurityPoliciesEditor.vue which resulted in save button being disabled at times


Aviv-Galmidi:
Run curieconf client Javascript (Jest) tests / coverage - Fails because of coverage. Known and will be fixed today

PR #770, opened 2022-03-20
Client improvements

Aviv-Galmidi:
Add decoding field - Content Filter Profile
Add content type - Content Filter Profile
Remove min risk level field - Content Filter Profile
Remove Content Filter Rule Groups from the documents dropdown
Remove validation of Regex input fields as it blocks valid rust regex
Fix bug where Security Policy had save button disabled when it shouldn't be
Fix bug where tags were added twice to ACL lists
Fix visual bug where long tags in list were breaking the view
Change icon in Rate Limit Rules Event of key name from regex to text, key value unaffected


Aviv-Galmidi:
Once again, Client unit tests fail for coverage

PR #772, opened 2022-03-21
Client improvements

Aviv-Galmidi:
Change CF Profile exclusion text from “Rule or group name” to “Content filter rules tags” and from “Ignore Content Filter” to “Ignore Content Filter Tags”
Change CF Profile exclusion from map to list of tags
Add check to prevent the user from deleting the last path in a Security Policy
Change Rate Limit Rule text from “A number of requests” to “Number of events”
Change order of documents in “Policies & Rules” dropdown to better reflect the flow of a request in the system


Aviv-Galmidi:
Once again, jest client tests fail for coverage

PR #788, opened 2022-03-28
sanity tests and signatures corrections

tzuryby:


tzuryby:
will re add it separatly, since Simon already submitted some of this earlier.

PR #809, opened 2022-04-19
Unified docker-compose file

bartavelle:
This patch modifies the default docker compose, so that the following
ports are exposed for the proxy backends:

 * 30081 - envoy echo server
 * 30082 - envoy juice shop
 * 30444 - envoy tls echo
 * 30445 - envoy tls juice
 * 31081 - nginx echo server
 * 31082 - nginx juice shop
 * 31444 - nginx tls echo
 * 31445 - nginx tls juice
 * 8080 - echo server
 * 3000 - juice shop


xavier-rbz (approved):
Looks good to me. Would you mind updating the documentation [here](https://docs.curiefense.io/reference/services-container-images) and [here](https://docs.curiefense.io/installation/deployment-first-steps/docker-compose)?

tzuryby:
@bartavelle in my diagram I forgot we have 2 ports, http and https....
should we have them both for Nginx as well, or shall we leave it http only?
In any case, I would rather have a unified use case (both, Envoy and Nginx, operating the same way)

bartavelle:
updated TLS ports

bartavelle:
How can I update the docs?

tzuryby (approved):

PR #812, opened 2022-04-19
manual docker build

yitzchake:
Add the ability to manually run the docker-push workflow.

Environment is set to "prod", which requires an approver.
Approver list can be found here:

https://github.com/curiefense/curiefense/settings/environments/470027045/edit

### DIFF
--- a/.github/workflows/curiefense-build-images.yml
+++ b/.github/workflows/curiefense-build-images.yml
@@ -1,6 +1,7 @@
 name: Publish docker images
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


xavier-rbz:
Thanks for the PR.
This mean approval is now required to run this when pushing to `main`, right? I don't think we want that.

yitzchake:
@xavier-rbz @tzuryby ,
PR has been updated. Build job works on merge without approval, as it did before.
I've created a separate workflow_dispatch job that does require approval.

> Thanks for the PR. This mean approval is now required to run this when pushing to `main`, right? I don't think we want that.

tzuryby (approved):

Issue #544, opened 2021-08-06
Weird behaviour when delete a profile of a urlmap


GabrielLasso:
**Describe the bug**
When I click to delete a profile, I don't know what happens, but it goes wrong (and I become unable to save, like in https://github.com/curiefense/curiefense/issues/543)
**To Reproduce**
1. Go to urlmaps
2. Fork some profiles
3. Delete a profile
4. Now the profiles are wrong

**Expected behavior**
Delete the selected profile

**Screenshots**
![Gravar_select-area_20210806114818](https://user-images.githubusercontent.com/12103372/128529044-a2409e4d-6289-4924-81a3-fc008154df0c.gif)

**Desktop (please complete the following information):**
 - OS: Manjaro
 - Browser: Opera

**Additional context**
Running on main. I deployed it yesterday

jdorfman:
Thanks @GabrielLasso for the detailed report and gif 🙌

I will send this over to @Aviv-Galmidi

spiros-p:
I just encountered the same bug. It seems that deleting a forked profile will _appear_ to assign its Match value back to an existing profile (overwriting whatever it had before), but the assignment isn't actually made.

Steps to illustrate this:
- After deleting the forked profile, do not save anything. 
- If more than one URL Map is defined, go to the pulldown on the upper right and switch the editor to a different URL Map, then return to the original one. The incorrect Match value is still displayed, which makes it appear as if the assignment was saved somehow.
- However, if you switch the editor to a different document type (e.g., Rate Limits), and then back to URL Maps, the incorrect value is no longer displayed.

GabrielLasso:
@spiros-p 
> However, if you switch the editor to a different document type (e.g., Rate Limits), and then back to URL Maps, the incorrect value is no longer displayed.

If I switch to a different document type and then back to URL Maps, the deletion will not be saved and the old profile will appear :/

spiros-p:
> If I switch to a different document type and then back to URL Maps, the deletion will not be saved and the old profile will appear :/

Yes, we're talking about slightly different scenarios. In my tests, I had forked the profile and immediately deleted it without ever saving it. 

I should have been more clear about this, and I've clarified my comment above. Sorry for the confusion.

PR #814, opened 2022-04-20
Cherry picked commits from merge branch that should be in main branch as well

Aviv-Galmidi:
Fix visual bug where buttons were not in the right place
Fix issue where deleting a URL Map entry would forcefully change another entry - Related https://github.com/curiefense/curiefense/issues/544
Fix Rate Limit being created with "empty" tag

### DIFF
--- a/curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
+++ b/curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
@@ -609,7 +609,7 @@ export default (Vue as VueConstructor<Vue & {
     },
     selectedBranch: {
       handler: function(val, oldVal) {
-        if (val && (!oldVal || !_.isEqual(val, oldVal))) {
+        if (val && !_.isEqual(val, oldVal)) {
           this.securityPoliciesDomainMatches()
         }
       },
--- a/curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
+++ b/curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
@@ -539,8 +539,8 @@ export default (Vue as VueConstructor<Vue & {
     },
 
     removeMapEntry(index: number) {
-      this.localDoc.map.splice(index, 1)
       this.changeSelectedMapEntry(-1)
+      this.localDoc.map.splice(index, 1)
     },
 
     referToRateLimit() {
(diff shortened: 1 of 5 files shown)


Aviv-Galmidi:
Failing Jest: Low coverage, no tests fail

curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
```diff
@@ -599,13 +598,23 @@ export default (Vue as VueConstructor<Vue & {
...
+            this.domainNames.push(val.match)
+          }
         }
       },
       immediate: true,
       deep: true,
     },
+    selectedBranch: {
+      handler: function(val, oldVal) {
+        if (val && (!oldVal || !_.isEqual(val, oldVal))) {
```
denis-reblaze, line 612: Excessive, `val && !_.isEqual(val, oldVal)` is enough

curiefense/ui/src/doc-editors/__tests__/SecurityPoliciesEditor.spec.ts
```diff
@@ -224,32 +225,33 @@ describe('SecurityPoliciesEditor.vue', () => {
...
+    selectedBranch = 'master'
     axiosGetSpy = jest.spyOn(axios, 'get').mockImplementation((path, config) => {
       if (!wrapper) {
         return Promise.resolve({data: []})
       }
       const branch = (wrapper.vm as any).selectedBranch
       if (path === `/conf/api/v2/configs/${branch}/d/aclprofiles/`) {
         if (config && config.headers && config.headers['x-fields'] === 'id, name') {
-          return Promise.resolve({data: _.map(aclDocs, (i) => _.pick(i, 'id', 'name'))})
+          return Promise.resolve({data: _.map(aclDocs, (i: ACLProfile) => _.pick(i, 'id', 'name'))})
```
denis-reblaze, line 236: Could we use more readable var names instead of `i`, in this case `doc` for instance?

denis-reblaze (changes requested):

curiefense/ui/src/doc-editors/SecurityPoliciesEditor.vue
Aviv-Galmidi: It was for very minor optimization to not call _.isEqual, really not worth it - removed

denis-reblaze (approved):

Issue #725, opened 2022-02-21
Curietasker docker image has missing python package


padziu:
**Describe the bug**
Curietasker image (curiefense/curietasker:main) has missing python package.
https://hub.docker.com/layers/curiefense/curietasker/main/images/sha256-72b5975502f208d93643ea0059d161cb8f8ae3b4be92aa9de89238f35e451d1f?context=explore

docker logs curietasker

`Determining whether the tasks namespace should be created...
The tasks namespace already exists.
Traceback (most recent call last):
  File "/usr/local/bin/curietasker", line 3, in <module>
    from curietasker import main
  File "/usr/local/lib/python3.7/dist-packages/curietasker/__init__.py", line 7, in <module>
    from . import tasker
  File "/usr/local/lib/python3.7/dist-packages/curietasker/tasker.py", line 7, in <module>
    from . import tasks_list_updates
  File "/usr/local/lib/python3.7/dist-packages/curietasker/tasks_list_updates.py", line 7, in <module>
    from jsonschema import validate, ValidationError
  File "/usr/local/lib/python3.7/dist-packages/jsonschema/__init__.py", line 21, in <module>
    from jsonschema._types import TypeChecker
  File "/usr/local/lib/python3.7/dist-packages/jsonschema/_types.py", line 6, in <module>
    from pyrsistent import pmap
ModuleNotFoundError: No module named 'pyrsistent'`

**To Reproduce**
docker-compose -f compose/docker-compose.yaml
docker logs curietasker

**Expected behavior**
Old image (curiefense/curietasker:latest) working fine.

Docker version 20.10.12, build e91ed57

chenstrace:
same issue

WaylonTian:
same too

Lastpixl:
Hello, this is indeed something that should be fixed. However, it can be ignored, and everything will still work properly.

xavier-rbz:
Hello, this is indeed something that should be fixed. However, it can be ignored, and everything will still work properly.

padziu:
@xavier-rbz This conatiner can't work properly beacuase it's restarting over and over again.
`afbf90d3d888   curiefense/curietasker:main                            "/usr/bin/dumb-init …"   5 days ago   Restarting (1) 20 seconds ago`

xavier-rbz:
Indeed, this container endlessly restarts, and needs fixing to avoid that.
[...]

PR #819, opened 2022-04-26
fix: ModuleNotFoundError: No module named 'pyrsistent' #725

MicroOps-cn:
see https://github.com/curiefense/curiefense/issues/725#issuecomment-1108105912


tzuryby:
curietasker was removed

PR #827, opened 2022-05-04
External processing patch

bartavelle:
This patch adds a new binary, and a new container, using the external
processing filter:

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto#external-processing-filter

This replaces the Lua filter. This increases latency when the system is
not loaded, but greatly increases throughput at scale.

There are still some rough edges related to configuration reloading:

 * because auto reloading is enabled, the code is a bit more complex
   than it should be, because it is not possible to hold a reference to
   the configuration. As a result, a lot of data needs to be copied and
   kept, instead of being a simple reference to the configuration.

 * it is now possible to get concurrent requests asking for
   configuration reloads, which get queued and might choke the system.

### DIFF
--- a/curiefense/curieproxy/lua/accesslog.lua
+++ b/curiefense/curieproxy/lua/accesslog.lua
@@ -9,12 +9,12 @@ local LOG_KEY = "request.info"
 local function get_log_table(request_map)
   -- handle is userData which is not serializable
   local entries = {
-    ["geo"]     = "geo",
-    ["headers"] = "headers",
-    ["cookies"] = "cookies",
-    ["args"]    = "arguments",
-    ["attrs"]   = "attributes",
-    ["tags"]    = "tags"
+    ["geo"]        = "geo",
+    ["headers"]    = "headers",
+    ["cookies"]    = "cookies",
+    ["args"]       = "arguments",
+    ["attributes"] = "attributes",
+    ["tags"]       = "tags"
   }
 
   local log_table = {}
--- a/curiefense/curieproxy/rust/Cargo.toml
+++ b/curiefense/curieproxy/rust/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "curiefense",
     "curiefense-lua",
     "curiefense-ffi",
+    "curiefense-externalprocessing",
 ]
 
 [profile.bench]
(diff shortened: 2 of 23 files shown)


xavier-rbz:
Some of the e2e tests don't pass, but it's not a case of total failure (that happen when a component is misconfigured, or when the Docker images have issues). Any idea what's wrong? I have skimmed the logs, nothing obvious.

I'd like to get performance numbers, let me know when it's ready to test on istio!

bartavelle:
I had to put back configuration auto reload to keep compatibility with the current mode of operation, and I saw some problems related to that, as with the async code you can now get concurrent configuration reload :/

bartavelle:
Also this is a POC, logs are not stored in elasticsearch for now, so the first test will always fail.

bartavelle:
I'll try to fix the logging issue tomorrow so that people can see something on elasticsearch, but it will not have the same format as the previous log, for now.

bartavelle:
And there is no grasshopper support!!!!

xavier-rbz:
I ran performance testing to compare this branch (7a8f6ae9fc34e5ae22d2de57920b3f9b3de6ebb7) with 1.13.2.
The `payloadsize_0kb` image is most interesting:
* as expected, CPU cost is higher: at 2000 RPS, ~2000m at worst for cf-1.13.2 vs ~2500m for all configs for external-processing. 
* latency is increased: a minimum of 2ms is observed with external-processing
* Max RPS in this setup is lower: latency spikes after ~2500 RPS for all configs with external-processing, whereas it stayed flat for default & denyall configs with 1.13.2

The denyall config graphs show latency increases dramatically with >4kB payloads, which did not happen with 1.13.2.

The default config graphs are inconclusive, as they show latency spikes in both versions.

bartavelle:
You should run the tests against configuration that use Redis (rate limiting / flow control), as this is that is supposed to be a showstopper in terms of scaling capability.

tzuryby:
@xavier-rbz 

the left (1.13.2)  and right (exProc) graphs seemed identical to my eyes at 01h43m.

what am I missing?

xavier-rbz:
@tzuryby indeed, I have uploaded the wrong files... Here are the correct ones:
compare-cf-1 13 2-with-external-processing--payloadsize_0kb
compare-cf-1 13 2-with-external-processing--config_denyall
compare-cf-1 13 2-with-external-processing--config_defaultconfig

PR #842, opened 2022-06-07
Global Filter fetch list "update now" fix

Aviv-Galmidi:
* Change fetch list to work with Curiefense-generated Global Filters
* Add test

### DIFF
--- a/curiefense/ui/src/doc-editors/GlobalFilterListEditor.vue
+++ b/curiefense/ui/src/doc-editors/GlobalFilterListEditor.vue
@@ -307,7 +307,7 @@ export default Vue.extend({
         let entries: GlobalFilterSectionEntry[]
         const convertedData = data as GlobalFilter
         if (convertedData?.rule?.sections?.length) {
-          this.localDoc.rule = data.rule
+          this.localDoc.rule = convertedData.rule
           this.localDoc.mdate = (new Date).toISOString()
           this.emitDocUpdate()
           return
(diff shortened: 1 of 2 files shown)


tzuryby:
@Aviv-Galmidi `rule.sections.length` is not a good test, since some lists may be empty as prt of the update.
I thought checking if the convertedData has the following properties

    id
    name
    action
    mdate
    source
    tags
    rule.sections

then it is most likely a valid structure, and then, sections overrides existing sections.
(even when empty)

curiefense/ui/src/doc-editors/GlobalFilterListEditor.vue
```diff
@@ -305,6 +305,13 @@ export default Vue.extend({
       RequestsUtils.sendRequest({methodName: 'GET', url: `tools/fetch?url=${url}`}).then((response: AxiosResponse) => {
         const data = response.data
         let entries: GlobalFilterSectionEntry[]
+        const convertedData = data as GlobalFilter
+        if (convertedData?.rule?.sections?.length) {
```
tzuryby, line 309: @Aviv-Galmidi `rule.sections.length` is not a good test, since some lists may be empty as part of the update.
I thought checking if the convertedData has the following properties

    id
    name
    action
    mdate
    source
    tags
    rule.sections

then it is most likely a valid structure, and then, sections overrides existing sections.
(even when empty)

Aviv-Galmidi: > some lists may be empty as part of the update

> sections overrides existing sections. (even when empty)

If the user wants to clear all sections we have a button for that "Clear all sections"

Using the "update now" button in order to clear the list sounds like an unwanted situation when trying to import a list of rules

I'm checking if there are any sections in the imported rule, not if any of the sections themselves are empty

If there are sections, even if some of them are empty, I would import the rules as a whole

PR #854, opened 2022-06-18
Add performance tests that use flow control & rate limits

xavier-rbz:
In this new performance test configuration, queries trigger both flow-control and ratelimit in tag-only mode, which should be the worst case for redis-related performance. This will help evaluate the performance of planned architecture changes where processing is done using an `ExternalProcessor`.

Sample performance results in the current `main` branch:
main15e0a8f881d9-payloadsize_0kb

### DIFF
--- a/deploy/deploy-gke.sh
+++ b/deploy/deploy-gke.sh
@@ -3,7 +3,7 @@
 # Pre-requisites:
 # * images built & pushed to the registry
 # * gcloud access is set up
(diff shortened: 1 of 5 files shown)


xavier-rbz:
Reviewers: I wrote this test configuration after getting advice from @bartavelle, but I might have messed it up. I'd be glad if you could cross-check it! Thanks

PR #861, opened 2022-06-29
bug fix: Same headers

fortianwei:
When openresty accept more than one headers with a same name, it put the headers' values to a `table`, which lead to a error because Rust code `headers` just accept `HashMap<String, String>`, not `HashMap<String, LuaTable>`.

### DIFF
--- a/curiefense/curieproxy/lua/session_nginx.lua
+++ b/curiefense/curieproxy/lua/session_nginx.lua
@@ -17,7 +17,15 @@ function session_rust_nginx.inspect(handle)
     end
 
     for k, v in pairs(rheaders) do
-        headers[k] = v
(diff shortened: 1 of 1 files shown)


bartavelle:
(also if you can fix the luacheck problem)

bartavelle:
But, that was a well spotted bug!

fortianwei:
> If you could intercalate spaces between the headers, and fix the DCO problem (`git commit -s`), that would be awesome!

Done with the DCO and luacheck problems :)

By the way, I'm not familar with envoy and wondering if the same issue exists in `session_envoy.lua`.

bartavelle:
I wonder too, will check that later :)

bartavelle (approved):

PR #864, opened 2022-07-07
Update Dockerfile

shubhshekhar09:
Latest does not exist changing to main.


xavier-rbz:
Hello,

Thanks for the interest in curiefense!
All images are built using [build-docker-images.sh](https://github.com/curiefense/curiefense/blob/main/curiefense/images/build-docker-images.sh) which overrides this correctly. Leaving `latest` here is a hint for the user that they are trying to build in an unsupported way.

PR #868, opened 2022-07-16
fix #866

### LINKED ISSUE #866: Curielogger unmarshalling error
**Describe the bug**
When using curiefense in our kubernetes cluster(AWS EKS), we find some requests' block  log not shown in kibana.
I think there maybe some errors  occurred  when `curielogger` working with the logs.

Then I find some error logs of currielogger:
`time="2022-07-16T09:24:51+08:00" level=error msg="Error occurred during unmarshalling. Error: json: cannot unmarshal array into Go struct field Response.response.headers of type string"`

**To Reproduce**
1. Testing some waf regex(like `rfi?..`)  in our service
2. error log can be found in curielogger's pod

**Expected behavior**
Every request that be blocked should be shown in kibana.

fortianwei:
Signed-off-by: Wei Tian <fortianwei@gmail.com>

As openresty make a `table` for same headers in both `request` and `response`, which leads to Rust/Golang  converting/unmarshalling error. I have updated the `session_nginx.lua` to make sure there is no `table` in headers.

@bartavelle

### DIFF
--- a/curiefense/curieproxy/lua/session_nginx.lua
+++ b/curiefense/curieproxy/lua/session_nginx.lua
@@ -6,27 +6,32 @@ local utils       = require "lua.nativeutils"
 local sfmt = string.format
 local custom_response = utils.nginx_custom_response
 
-function session_rust_nginx.inspect(handle)
-    local ip_str = handle.var.remote_addr
-
+local function make_safe_headers(rheaders)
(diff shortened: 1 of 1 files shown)


bartavelle:
Thanks a lot!

bartavelle (approved):

PR #924, opened 2022-10-04
New e2e tests

bartavelle:
Replace the e2e tests so that it uses the lua test suite.

Currently needs:

  * support the flow control tests
  * add more things to test


tzuryby (approved):

xavier-rbz:
This removes bitflip tests, most of which would be best done in the curieproxy unit tests. If we remove them from e2e tests, we won't see bugs coming from envoy, or bugs originating from way envoy interacts with lua+rust that might be different from what the lua+rust unit tests do.

bartavelle:
> This removes bitflip tests, most of which would be best done in the curieproxy unit tests. If we remove them from e2e tests, we won't see bugs coming from envoy, or bugs originating from way envoy interacts with lua+rust that might be different from what the lua+rust unit tests do.

This is WIP, I would like to make it work like it worked before (with pytest), but checked if that was feasible with "real requests" first.
I would like to add the bitflip test, that part should not be too hard. However, I don't know how to make it behave like a pytest suite for now!

bartavelle:
Managed to convert back to the pytest format! next: re-enable bit flip tests

bartavelle:
Bit flip tests are back, but they take a long time to run. Compared to the rust compilation step, it's not that long, @xavier-rbz what do you think of adding a probability to run them to save some time?

tzuryby (approved):

xavier-rbz:
> @xavier-rbz what do you think of adding a probability to run them to save some time?
Thanks for adding them back!
Running them for a random xx% of test runs would make tests undeterministic, I'm not convinced. To make tests faster, we could run these on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) instead? Let's think about it.

PR #1006, opened 2022-10-24
add prometheus client dependencies to setup

shaisgv:


tzuryby:
where is this coming from?
not sure I am willing to "play" with curieconf at this moment, unless it is a necessity

shaisgv:
@tzuryby, this was requested by Evegeniy,.
@Freakachoo , can you please answer Zuri.

Freakachoo:
@tzuryby we're using Prometheus to export flask metrics. 
Here:
https://github.com/curiefense/curiefense/blob/main/curiefense/curieconf/server/curieconf/confserver/__init__.py#L9
But this module wasn't added to dependencies. So, here we are just adding it.

curiefense/curieconf/server/setup.py
```diff
@@ -38,6 +38,7 @@
         "fasteners",
         "jsonpath-ng==1.5.3",
         "pydash==5.0.2",
+        "prometheus-flask-exporter==0.20.3",
```
Freakachoo, line 41: @shaisgv seems like it has different indentation

PR #1009, opened 2022-10-25
feat: partial support of ipinfo geoip and autotag

Rileran:
## Description

Add support for ipinfo database ip intelligence.

## Features

- Allow usage of both maxmind and ipinfo, with ipinfo as primary source of truth.
- Allow disabling maxmind alltogether.
- Fallback to maxmind if ipinfo can't be loaded (current behaviour)
- Add new security traits from ipinfo (vpn, tor, relay, hosting, service_name)

As ipinfo and maxmind does not have a 1-1 mapping of all fields, the data may be different if using one of the two systems.

## Difference with maxmind

- Subregion information is not available (though postal code may be)
- No continent information (neither continent name nor continent code)
- No country name (but country iso is available, and ipinfo provides json mapping https://github.com/ipinfo/rust/blob/master/src/countries.json)
- No "in_eu" (but ipinfo provides a json set https://github.com/ipinfo/rust/blob/master/src/eu.json)


curiefense/curieproxy/rust/curiefense/src/tagging.rs
```diff
@@ -238,15 +238,29 @@ pub fn tag_request(
...
+        tags.insert("ipinfo-tor", Location::Ip)
+    }
+    if let Some(true) = rinfo.rinfo.geoip.is_relay {
+        tags.insert("ipinfo-relay", Location::Ip)
+    }
+    if let Some(true) = rinfo.rinfo.geoip.is_hosting {
+        tags.insert("ipinfo-hosting", Location::Ip)
+    }
+    if let Some(privacy_service) = rinfo.rinfo.geoip.privacy_service.as_deref() {
+        tags.insert_qualified("ipinfo-privacy_service", privacy_service, Location::Ip)
```
bartavelle, line 260: perhaps `ipinfo-privacy-service`

bartavelle (approved):

tzuryby:
this should not be pushed yet.
the final would be optional conf (ENV variable) to pick the provider (maxmind/ipinfo) and the file names which seems like we will need to do some work on the curieconf in parallel 
see https://github.com/curiefense/curiefense/blob/a6954ff4b7b0af2e80578b3bd13d5291196f270b/curiefense/curieconf/utils/curieconf/utils/__init__.py#L22

SimonMarechal:
Currently it defaults on maxmind, but can be configured to use ipinfo or both at the same time with environment variables.

Rileran:
This PR is currently missing integration for [ipinfo ASN API}(https://ipinfo.io/products/asn-api) (in particular to get the `route` attribute for the `network` field) because we are missing the database from ipinfo.

Maybe @tzuryby has more information on the subject.

tzuryby:
@Rileran 

 1. we do have the files ([ipinfo ASN API](https://ipinfo.io/products/asn-api)), while they cannot be included in the "public"repository. will share with @bartavelle soon directly.
 2. There is no need to support "both, while x is primary", rather ot support either one. Meaning, at config, a user pick the "vendor", provide paths to db files, and our system will lookup using that only DB.
 3.

Rileran:
@tzuryby Could you please send the asn file to @bartavelle so that I can finish this feature ?

PR #1070, opened 2022-11-23
Update build-docker-images.sh

tzuryby:
@xavier-rbz many time we clone with `git clone --depth=1` jsut to get the latest.
in those cases, the command
```
git describe --tag --long --dirty
```
produces:
```
fatal: No names found, cannot describe anything.
```

to avoid this, I suggest the following changes


xavier-rbz (approved):
LGTM

PR #1086, opened 2022-11-24
Refacto: action type

Rileran:
## Description

Separate rust Action and Description representation to match the actual response


tzuryby:
@bartavelle what's the status of this ?

PR #1130, opened 2022-12-14
feat: basic support for config file reloading

Rileran:


bartavelle:
Note that `cargo test` does not compile.

tzuryby (approved):

PR #1161, opened 2023-02-02
Take earliest interval in metrics

Freakachoo:
Because metrics are getting a list of 2 objects: full metrics for the previous interval, and partial for the current, we need to take only the previous interval.


xavier-rbz:
Would you mind running black, so that CI becomes green on this point? You might want to review & merge #1163 that fixes invalid formatting on code that is not modified by this PR.

Generally, I'm happier when what we merge is greenlit by CI. That prevents regressions, and tiring debugging sessions later.

PR #1192, opened 2023-03-14
using echo.reblaze.com for the echo server

yitzchake:
Update the echoserver for v5 reblaze planets.
They should use echo.reblaze.com, instead of a pod.

The current echo server pod can only receive http requests, and echo.reblaze.com is already configured for https requests as well.

It would also mean fewer resources wasted on a pod that's only used on planet startup

Planet with new reblaze echoserver built and tested here:
- https://github.com/reblaze/rbz-infra-v5-planets/actions/runs/4403637369


yanushg (approved):

tzuryby:
Not approving.

PR #1200, opened 2023-03-28
Rr2984

yoavkatzman:
a fix to https://reblaze.atlassian.net/browse/RR-2984
maintaining "system" with it's original fields (extra can be added)

### DIFF
--- a/curiefense/curieconf/server/curieconf/confserver/v3/api.py
+++ b/curiefense/curieconf/server/curieconf/confserver/v3/api.py
@@ -935,9 +935,12 @@ async def ns_resource_get(nsname: str, request: Request):
 
 
 @router.post("/db/{nsname}/", tags=[Tags.db])
-async def ns_resource_post(nsname: str, db: DB, request: Request):
+async def ns_resource_post(nsname: str, request: Request):
     """Create a non-existing namespace from data"""
     _db = await request.json()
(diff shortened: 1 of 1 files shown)


curiefense/curieconf/server/curieconf/confserver/v3/api.py
```diff
@@ -934,25 +935,32 @@ async def ns_resource_get(nsname: str, request: Request):
 
 
 @router.post("/db/{nsname}/", tags=[Tags.db])
-async def ns_resource_post(nsname: str, db: DB, request: Request):
+async def ns_resource_post(nsname: str, request: Request):
     """Create a non-existing namespace from data"""
     _db = await request.json()
+    if nsname == "system" and set.intersection(set(_db.keys()),set(bootsrap_system_keys)) != set(bootsrap_system_keys):
```
Freakachoo, line 941: “system” should go as a constant and probably be a list. But I’m not sure that we really need this at the moment.

Freakachoo, line 941: IMO it's better to break it into 2 validation functions. One will check "system", another one all the rest. And use them as a Dependency. Like in the example:
```
dependencies=[Depends(verify_providers([PROVIDER1, PROVIDER2]))],
```
So those functions could be reused in many handlers

Freakachoo (dismissed):

PR #1201, opened 2023-03-29
initial bootstrap fix

yoavkatzman:
handling bootstrap errors to eliminate the problem of confserver not starting on creation.
considering the behaviour is that re running bootstrap always fixed the problem to this day- this fix runs bootstrap up to 3 times in case of failures.

### DIFF
--- a/curiefense/images/confserver/init/entrypoint.sh
+++ b/curiefense/images/confserver/init/entrypoint.sh
@@ -1,12 +1,19 @@
 #!/bin/bash
-
 # make sure git trusts the persistennt directories.
+
 git config --global --add safe.directory /cf-persistent-config/confdb
 git config --global --add safe.directory /cf-persistent-config/bootstrap-repo
 
 if [ "$INIT_GIT_ON_STARTUP" = "yes" ]; then
 	# used when running with docker-compose, which does not have initContainers or similar features
-	/bootstrap/bootstrap_config.sh
+	for i in {1..3}
+  do
+	  /bootstrap/bootstrap_config.sh
+	  if [ $? -eq 0 ]; then
+	    echo "error - entrypoint.sh bootstrap_config failed, attempt number $i, trying again"
+      break
+    fi
+  done
 fi
 
 /bootstrap/initial-bucket-export.sh &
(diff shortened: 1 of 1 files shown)


curiefense/images/confserver/init/entrypoint.sh
```diff
@@ -1,12 +1,19 @@
...
 	# used when running with docker-compose, which does not have initContainers or similar features
-	/bootstrap/bootstrap_config.sh
+	for i in {1..3}
+  do
+	  if /bootstrap/bootstrap_config.sh; then
+      break
+    fi
+    echo "error - entrypoint.sh bootstrap_config failed, attempt number $i, trying again"
+
+  done
```
Freakachoo, line 16: Something with indentation here

curiefense/images/confserver/init/entrypoint.sh
```diff
@@ -1,12 +1,19 @@
...
-
 # make sure git trusts the persistennt directories.
+
 git config --global --add safe.directory /cf-persistent-config/confdb
 git config --global --add safe.directory /cf-persistent-config/bootstrap-repo
 
 if [ "$INIT_GIT_ON_STARTUP" = "yes" ]; then
 	# used when running with docker-compose, which does not have initContainers or similar features
-	/bootstrap/bootstrap_config.sh
+	for i in {1..3}
```
Freakachoo, line 9: Could you leave a comment here WHY we repeat it a few times. That it happens from time to time, etc...

yoavkatzman: fixed, for some reason it doesn't show on the PR (but you can see it when you click view file)

curiefense/images/confserver/init/entrypoint.sh
```diff
@@ -1,12 +1,19 @@
...
 git config --global --add safe.directory /cf-persistent-config/confdb
 git config --global --add safe.directory /cf-persistent-config/bootstrap-repo
 
 if [ "$INIT_GIT_ON_STARTUP" = "yes" ]; then
-	# used when running with docker-compose, which does not have initContainers or similar features
-	/bootstrap/bootstrap_config.sh
+  # used when running with docker-compose, which does not have initContainers or similar features
+  # due to unexpected failures, bootstrap will run up to 3 times to better eliminate these failures (usually fixed after one failure)
+  for i in {1..3}; do
+    if /bootstrap/bootstrap_config.sh; then
```
Freakachoo, line 11: Let's add some sleep for a few seconds in case it is happening because of some race condition

Freakachoo (approved):

olgasobolevsky (approved):

Issue #1065, opened 2022-11-22
Content filter profile, Security policy - action tags in log when profile is inactive


tzuryby:
When content filter profile is inactive, we show action tags and response status in logs. 
When ACL profile is inactive, we don't show action tags and response status in logs. 
Same behavior is expected in both cases.

**Steps to reproduce:**
1. Security policies -> set the content filter profile to be inactive (unchecked)
![sp example](https://user-images.githubusercontent.com/56600129/198865961-d84031d4-d77c-418d-a0ac-5dbf58d5f9f1.png)
2. Send request to activate content filter

**Actual result:** 
Response in terminal: 200
Response in logs: 473, action: content-filter-block
![curl](https://user-images.githubusercontent.com/56600129/198866018-e83db523-876f-44af-80cb-dc64211d2689.png)
![log example ](https://user-images.githubusercontent.com/56600129/198866022-ea305cab-2d97-4f68-b8f7-4eda4043877a.png)
![log example2](https://user-images.githubusercontent.com/56600129/198866027-83ad955f-c10c-45dc-aa33-c7a65b28bc8d.png)

**Expected result:**
Same response status in terminal and in log, no action tags

### PREVIOUS COMMETNS

by @bartavelle 

> That is a long-standing problem, but now we have extra data in the log file. 
> It could be possible to remove this information and rely on the *_trigger fields instead to know what would happen 
> if the content filter was active. Should I do that?

tzuryby:
@bartavelle not sure I fully understand the question. let's discuss this.

@olgasobolevsky - issue moved here

bartavelle:
We will fix the status issue, but what about the tags? Should they be added when the acl / content filter is inactive, or not?

PR #1228, opened 2023-05-22
Fix response code in log for inactive content filter

bmwant:
### Description

Set status code from the content filter only when it is active.
In the other case respect status code from the upstream and serialize it to the logs.
Add `blocked` method to reuse in logs and in status code block logic.

Addressing this issue https://github.com/curiefense/curiefense/issues/1065

### Testing
* `473` when content filter is active
473_filter_active

* `200` when content filter is inactive
200_filter_inactive

*  Logs in Kibana shows correct response code for both cases
response_logs_correct_status


curiefense/curieproxy/rust/curiefense/src/interface/mod.rs
```diff
@@ -512,14 +527,8 @@ pub fn jsonlog_rinfo(
...
-    //blocked
-    let mut blocked = false;
-    for r in &dec.reasons {
-        if !(matches!(r.action, RawActionType::Monitor) || matches!(r.action, RawActionType::Skip)) {
-            blocked = true;
-            break;
-        }
-    }
+    // blocked
+    let blocked = dec.blocked();
```
arbelaxRbz, line 531: The logic of the "blocked" log param was updated a bit in branch rr_3127_fix_blocked_challenge_logs
(See https://github.com/curiefense/curiefense/blob/c7854552a24cccc44dca01d3e7db97600f324116/curiefense/curieproxy/rust/curiefense/src/interface/mod.rs#L517)
So this part might have to be adapted after that pr is merged next week

bmwant:
Testing tag `action:content-filter-block` is not present when inactive (and present when enabled)
Screenshot 2023-06-01 at 6 38 11 PM

olgasobolevsky (approved):

arbelaxRbz (approved):

Freakachoo (approved):

